PR #537, opened 2025-12-09
[fix] remove '@' from mcp tool name

weizjajj:
deepseek not support @ in tool name


hxbszyy:
Closing this because it does not actually remove @ by default: the new separator still defaults to @, is not validated, and has no tests for registration, lookup, collision handling, or model-safe naming. MCP tool names need one canonical sanitization/alias contract across all providers, not a per-toolkit free-form separator.

PR #547, opened 2026-01-13
docs: add Dockerfile and document local development setup

Nuteer:
## What was added

- Added a production-ready Dockerfile for building and running agentUniverse.
- Documented local development setup using Poetry.
- Documented Docker build and run instructions.
- Added a basic sanity check command.

## Why

This improves contributor onboarding and makes it easier to:
- Build the project locally.
- Validate environment setup.
- Run the project inside Docker without manual dependency installation.

## Notes

The Docker setup intentionally does not run the full upstream test suite automatically,
since several tests depend on external services, plugins, and runtime configuration.

This PR focuses on build reproducibility and developer experience.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM python:3.10-slim
+
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR /app
+
+# System dependencies
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    curl \
+    git \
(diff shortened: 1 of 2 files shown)


hxbszyy:
Closing this because the Docker image is not a reliable development or runtime contract in its current form. It runs Poetry root installation before the package source and README are copied, has no lock file or .dockerignore, copies the entire checkout into the image, keeps compilers and Git in a root runtime image, and the entry point only imports the package. A fresh Docker contribution should start from current master with a reproducible dependency layer, a non-root runtime stage, an explicit runnable target, and a build/smoke test.

PR #553, opened 2026-02-25
fix: replace 9 bare excepts with except Exception

haosenwang1018:
Replace 9 bare `except:` with `except Exception:` across 9 files. Bare `except:` catches `KeyboardInterrupt` and `SystemExit`, masking real errors.


hxbszyy:
Closing this sweep. Replacing bare except clauses mechanically across unrelated subsystems does not address why each exception is swallowed, adds no behavioral tests, and the branch has had no follow-up while those files have continued to change. Exception handling should be fixed in focused PRs that preserve cancellation and shutdown behavior and either recover deliberately or log enough context.

PR #556, opened 2026-03-02
feat: add version constant and get_version() helper

haosenwang1018:
Add __version__ constant and get_version() function.


hxbszyy:
Closing this version as well. It carries the unrelated nine-file exception sweep from #553, while the actual version helper hard-codes 0.0.19 separately from pyproject.toml and will drift on the next release. If a public runtime version API is needed, please derive it from installed package metadata and submit that as a small current-master PR with source and wheel tests.

PR #560, opened 2026-03-28
fix: add default value to audio_url pop in agent.py

### LINKED ISSUE #540: 🐞 [Bug] <同一个agent中，使用多个prompt模板，第一次之后组装prompt会报错>
### Version

version:0.0.18

### System

mac

### Actions

_No response_

### Problem

Image

这里agent_input.pop('audio_url')使用了pop，会移出agent_input中这个key，并且没有带默认值，第二次调用则会抛出异常

### Expected

pop带上默认值可以修复。
但是为什么用pop，理论上不应该删掉agent_input中的key吧。

### reproduce

_No response_

yang1002378395-cmyk:
### Problem
When using multiple prompt templates in the same agent, the second call fails because `agent_input.pop('audio_url')` doesn't have a default value. The key gets removed on the first call, causing a `KeyError` on subsequent calls.

### Solution
Changed `agent_input.pop('audio_url')` to `agent_input.pop('audio_url', '')` to match the pattern already used for `image_urls` on line 466.

### Changes
- Line 468 in `agentuniverse/agent/agent.py`: Added default value to pop() call

Fixes #540

### DIFF
--- a/agentuniverse/agent/agent.py
+++ b/agentuniverse/agent/agent.py
@@ -465,7 +465,7 @@ class Agent(ComponentBase, ABC):
         if image_urls:
             chat_prompt.generate_image_prompt(image_urls)
 
-        audio_url: str = agent_input.pop('audio_url') or ''
+        audio_url: str = agent_input.pop('audio_url', '') or ''
         if audio_url:
             chat_prompt.generate_audio_prompt(audio_url)
         return chat_prompt


hxbszyy:
Closing this because the bug has already been resolved on current master in a safer form: the audio path now reads audio_url with a default instead of destructively popping it. This branch conflicts with the current implementation and no longer adds behavior.

PR #563, opened 2026-03-28
fix: add null check for async_client in as_langchain methods

yang1002378395-cmyk:
## Problem

When using `as_langchain()` method before calling async methods, `async_client` is `None`, causing:

```
AttributeError: 'NoneType' object has no attribute 'embeddings'
```

Related to #188

## Solution

- Add null check for `async_client` in both `OpenAIEmbedding` and `AzureOpenAIEmbedding`
- Only include `async_client` in kwargs when it's not None
- Maintain backward compatibility

## Changes

- `agentuniverse/agent/action/knowledge/embedding/openai_embedding.py`
- `agentuniverse/agent/action/knowledge/embedding/azureopenai_embedding.py`

## Testing

- Syntax check passed
- No breaking changes to existing API

### DIFF
--- a/agentuniverse/agent/action/knowledge/embedding/azureopenai_embedding.py
+++ b/agentuniverse/agent/action/knowledge/embedding/azureopenai_embedding.py
@@ -102,7 +102,14 @@ class AzureOpenAIEmbedding(Embedding):
         self._initialize_clients()
 
         from langchain_community.embeddings.azure_openai import AzureOpenAIEmbeddings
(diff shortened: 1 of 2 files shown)


hxbszyy:
Closing this because current master has already changed the client lifecycle. The OpenAI-style adapter now creates sync and async clients through dedicated lazy initializers, and the Azure adapter initializes both clients before building its LangChain wrapper. The added null guards are therefore redundant on the current paths, while this branch conflicts with the implementation it was written against.

PR #584, opened 2026-06-21
fix(agent_serve): allow list and dict types for RequestDO.query

wustwyh:
The RequestDO Pydantic model previously constrained query to str, causing a validation error when service requests passed list or dict parameters (e.g. input: []). This change relaxes the type to Union[str, list, dict] so that non-string request payloads are accepted.

Adds unit tests covering string, list, and dict query values, plus a model_dump round-trip assertion.

### DIFF
--- a/agentuniverse/agent_serve/web/dal/entity/request_do.py
+++ b/agentuniverse/agent_serve/web/dal/entity/request_do.py
@@ -7,7 +7,7 @@
 # @FileName: request_do.py
 
 import datetime
-from typing import Optional
+from typing import Optional, Union
 
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
I am going to hold this one for now.

The model-level change is in the right direction, and the new `RequestDO` tests pass. I also ran the PR-owned test file in the conda `agentUniverse` environment: Python 3.10.16, pytest 7.4.4. `py_compile`, `git diff --check`, and `tests/test_agentuniverse/unit/agent_serve/test_request_do.py` all pass.

The blocker is the saved request path. `RequestLibrary` still maps `query` to a SQLAlchemy `Text` column, and `add_request()` passes `request_do.model_dump()` straight into the ORM. With this PR applied, `RequestDO` accepts list/dict values, but saving them still fails:

- `str`: saved
- `list`: `sqlite3.InterfaceError: Error binding parameter ... unsupported type`
- `dict`: `sqlite3.InterfaceError: Error binding parameter ... unsupported type`

So the original validation error is fixed, but saved requests with non-string `query` still break one step later. Please either persist `query` as JSON, or serialize/deserialize non-string query values consistently in `RequestLibrary`, and add a regression test that covers `RequestLibrary.add_request()` for list/dict queries. Once that path works, I can take another look.

Issue #250, opened 2025-01-16
Enhancements for more LLM components/LLM组件接入


LandJerry:
======================== English version ==============================
The expected LLM model components are as follows:

* Azure OpenAI
* AWS
* Google
* huggingface_hub
* vllm
* Other LLM components not included in the aU project.

======================== 中文版 ==============================
期待您优先接入的LLM模型组件如下：
* Azure OpenAI
* AWS
* Google
* huggingface_hub
* vllm
* 其他项目中未包含的同类组件

Yuky-499:
I'd like to contribute two independent follow-ups to the existing AWS Bedrock components (LLM #250, embedding #251).

1. **Optional dependencies and offline tests**: declare an `aws_ext` extra for boto3, retain lazy imports, make unit tests use injected clients without boto3/AWS credentials, and gate live AWS tests behind an explicit environment flag. Include English/Chinese installation and testing instructions. No asynchronous behavior changes in this PR.

2. **Non-blocking async APIs**: offload synchronous boto3 work for LLM calls and embeddings using Python 3.10's `asyncio.to_thread`. For streaming, move both request creation and iterator reads off the event loop, preserve ordering, avoid unbounded prefetch, and serialize reading/closing. Cancellation must propagate without waiting on blocking network I/O; an in-flight SDK operation cannot be forcibly interrupted, so cleanup must occur once it finishes (subject to the SDK timeout). Avoid boto3's shared default session in concurrent creation. Add deterministic offline tests for event-loop responsiveness, errors, cancellation, empty streams, early close and resource cleanup.

Both branches will start from current master. This does not include the separate empty non-streaming response parsing change proposed in #953, nor a LangChain migration or an aioboto3 dependency. I'll link the PRs here for review; this is a proposed design, not an assumption of maintainer approval.

Yuky-499:
Implementation follow-up: the two independent PRs are now open for review: #2035 (optional AWS SDK dependency and offline/live-test contract) and #2036 (non-blocking native async LLM/embedding calls, streaming ownership and cancellation cleanup). Both start from master and combine without conflicts. Combined offline validation: 76 passed without boto3; 79 passed with boto3 1.34.131 and 1.43.88, with live AWS tests explicitly skipped. No real AWS requests were made. The async PR includes an English/Chinese design document explaining lazy stream opening, per-stream worker ownership, public tracing cleanup, and the limits of cancelling synchronous SDK work.

[...]

PR #597, opened 2026-07-06
feat: add Azure OpenAI LLM component

itxaiohanglover:
## Summary
- Addresses task #250: Adds Azure OpenAI LLM component
- Creates `AzureOpenAILLM` class extending `OpenAIStyleLLM` with `AzureOpenAI`/`AsyncAzureOpenAI` client support
- Supports configurable `api_version`, Azure endpoint (via `api_base`), proxy, and context length lookup for common Azure OpenAI deployments (gpt-35-turbo, gpt-4, gpt-4o, o1, o3-mini, etc.)

## Usage
Configure in YAML:
```yaml
name: 'default_azure_openai_llm'
description: 'default azure openai llm with spi'
model_name: 'gpt-4o-mini'  # Azure deployment name
max_tokens: 1000
api_key: '${AZURE_OPENAI_API_KEY}'
api_base: '${AZURE_OPENAI_ENDPOINT}'
api_version: '2024-10-21'
metadata:
  type: 'LLM'
  module: 'agentuniverse.llm.default.azure_openai_llm'
  class: 'AzureOpenAILLM'
```

Environment variables:
- `AZURE_OPENAI_API_KEY` - Azure OpenAI API key
- `AZURE_OPENAI_ENDPOINT` - Azure OpenAI endpoint URL
- `AZURE_OPENAI_API_VERSION` - API version (default: 2024-10-21)
- `AZURE_OPENAI_PROXY` - Optional proxy

## Test plan
- [x] Verified Python syntax compiles correctly
- [x] Component follows the same pattern as existing LLM components (GeminiOpenAIStyleLLM, DefaultOpenAILLM)

### DIFF
--- a/agentuniverse/llm/default/azure_openai_llm.py
+++ b/agentuniverse/llm/default/azure_openai_llm.py
@@ -1,0 +1,129 @@
+# !/usr/bin/env python3
+# -*- coding:utf-8 -*-
+
+# @Time    : 2026/07/06
+# @FileName: azure_openai_llm.py
+
+from typing import Any, Optional, Union, Iterator, AsyncIterator
+
+import httpx
+from pydantic import Field
+
+from agentuniverse.base.util.env_util import get_from_env
+from agentuniverse.llm.llm_output import LLMOutput
+from agentuniverse.llm.openai_style_llm import OpenAIStyleLLM
+
+AZURE_OPENAI_MAX_CONTEXT_LENGTH = {
+    "gpt-35-turbo": 4096,
+    "gpt-35-turbo-16k": 16384,
+    "gpt-35-turbo-1106": 16384,
+    "gpt-35-turbo-0125": 16384,
+    "gpt-4": 8192,
+    "gpt-4-32k": 32768,
+    "gpt-4-0613": 8192,
+    "gpt-4-1106-preview": 128000,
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
I like the direction of this PR: Azure OpenAI is a useful provider for agentUniverse and fits the existing LLM integration model. I cannot merge the current implementation yet, though.

The issue is that `AzureOpenAILLM` builds an `AzureOpenAI` client, but then delegates to `OpenAIStyleLLM._call()`. That parent method resets `self.client.base_url` to `self.api_base` before each request. With the Azure client, the initialized base URL is `https://<resource>.openai.azure.com/openai/`, but the inherited call path changes it back to `https://<resource>.openai.azure.com`. That bypasses part of the Azure client URL construction and is very likely to send requests to the wrong path.

I reproduced the state transition locally with a dummy endpoint: `_new_client()` produced `https://example-resource.openai.azure.com/openai/`, then the inherited `_call()` logic would reset it to `https://example-resource.openai.azure.com`.

Please adjust the Azure implementation so the parent call path does not mutate the Azure client's `base_url` this way, and add a small regression test or mocked-client smoke test around the Azure client setup/call path. Once that is covered, this looks like a reasonable provider addition to revisit.

PR #598, opened 2026-07-06
feat: add HuggingFace Hub LLM component

itxaiohanglover:
## Summary

Adds `HuggingFaceHubLLM` — a new LLM component that uses `huggingface_hub.InferenceClient` to interact with models deployed on the HuggingFace Inference API or dedicated TGI endpoints.

### Features
- Sync and async chat completions via `InferenceClient` / `AsyncInferenceClient`
- Streaming support
- Configurable temperature, max_tokens, model selection
- Context length lookup for common HF models (Llama, Mistral, Qwen, Gemma)
- YAML config with env var support (`HF_TOKEN`, `HUGGINGFACEHUB_API_TOKEN`, `HUGGINGFACEHUB_API_BASE`)

### Files Changed
- `agentuniverse/llm/default/huggingface_hub_llm.py` — Main component
- `agentuniverse/llm/default/huggingface_hub_llm.yaml` — Default config
- `tests/test_agentuniverse/unit/llm/test_huggingface_hub_llm.py` — 12 unit tests

### Implementation Details
- **Async client**: Uses separate `AsyncInferenceClient` instance (not shared sync client)
- **API correctness**: Uses `client.chat_completion()` — the correct HF Hub API, not `chat_completions.create()` (OpenAI style)
- **Code quality**: Extracted `_build_kwargs()` and `_to_raw()` helpers to reduce duplication
- **Token counting**: Implements `get_num_tokens()` using tiktoken with `cl100k_base` fallback
- **Tests**: 12 mocked unit tests covering sync/async calls, streaming, client caching, and config

### Checklist
- [x] DCO signed
- [x] Unit tests pass (12/12)
- [x] No breaking changes (new component only)

### DIFF
--- a/agentuniverse/llm/default/huggingface_hub_llm.py
+++ b/agentuniverse/llm/default/huggingface_hub_llm.py
@@ -1,0 +1,225 @@
+# !/usr/bin/env python3
+# -*- coding:utf-8 -*-
+
+# @Time    : 2026/07/06
+# @FileName: huggingface_hub_llm.py
+
+from typing import Any, Optional, Union, Iterator, AsyncIterator
+
+from pydantic import Field
+
+from agentuniverse.base.util.env_util import get_from_env
+from agentuniverse.llm.llm import LLM, LLMOutput
+
+try:
+    import tiktoken
+except ImportError:
+    tiktoken = None
+
+HUGGINGFACE_HUB_MAX_CONTEXT_LENGTH = {
+    "meta-llama/Llama-3.1-8B-Instruct": 128072,
+    "meta-llama/Llama-3.1-70B-Instruct": 128072,
+    "meta-llama/Llama-3.2-1B-Instruct": 128072,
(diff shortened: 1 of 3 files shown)


hxbszyy (changes requested):
This is a useful direction for agentUniverse: a HuggingFace Hub/TGI-backed LLM provider fits the existing provider integration story. I am not comfortable merging it yet because the new runtime dependency is not declared by the project.

The component imports `huggingface_hub.InferenceClient` / `AsyncInferenceClient`, and the tests patch those classes, but `huggingface_hub` / `huggingface-hub` is not listed in `pyproject.toml` or an optional extra. The tests pass in my current environment only because that package happens to already be installed there. A fresh install from the project metadata would not be able to use this built-in LLM component.

Please add the dependency in the appropriate place, or make this an explicit optional extra with a clear import-time error telling users what to install. After that, the current unit tests, `py_compile`, and diff hygiene are already looking good.

PR #601, opened 2026-07-06
fix: add input-level sandbox to PythonREPLTool

### LINKED ISSUE #570: PythonREPLTool executes arbitrary code without sandboxing
## Bug Description

`PythonREPLTool` in `agentuniverse/agent/action/tool/common_tool/python_repl.py` passes its `input` argument directly to `langchain_community.utilities.PythonREPL.run()`, which calls Python's `exec()` with no sandboxing, subprocess isolation, or resource limits. Any agent instruction that reaches this tool executes on the host operating system.

## Location

`agentuniverse/agent/action/tool/common_tool/python_repl.py:26`

## Reproduction

1. Equip an agent with `PythonREPLTool`.
2. Send any input that causes the agent to invoke the tool — or trigger prompt injection via untrusted web/document content that the agent processes.
3. The tool calls `self.client.run(input)`, running `exec(input)` on the host.

Example payload: `__import__('os').system('id')` or `open('/etc/shadow').read()`.

## Impact

In any agentic pipeline exposed to user input, prompt injection escalates directly to arbitrary code execution on the server. The tool provides no path restriction, resource cap, or subprocess isolation. Secrets, credentials, and the full host filesystem are accessible to the injected payload.

## Suggested Fix

- Wrap execution in a subprocess with a restricted environment (`subprocess.run` with a dedicated low-privilege user and no network).
- Or add a prominent security warning in the docstring and README, requiring explicit opt-in from integrators who understand the risk.
- Remove the tool from any default agent configuration that processes user-supplied content.

---

itxaiohanglover:
## Summary

Fixes #570

Adds a lightweight input-level sandbox to `PythonREPLTool` that blocks dangerous patterns (e.g. `__import__`, `os.system`, `subprocess`, `eval`) before code execution.

### Changes

| File | Description |
|------|-------------|
| `python_repl.py` | Add `sandbox_enabled` and `blocked_patterns` fields, `_check_safety()` method |
| `test_python_repl_tool.py` | 12 unit tests covering blocking behavior and safe code execution |

### Blocked Patterns (default)
- `__import__()`, `import os/subprocess/shutil/socket/ctypes/sys`
- `os.system()`, `os.popen()`, `subprocess.*`
- `eval()`, `exec()`, `compile()`
- `open()` with absolute or `~` paths
- `getattr()`, `setattr()`, `delattr()`
- `globals()`, `locals()`

### Configuration
```yaml
sandbox_enabled: true  # default: true, set false to disable
blocked_patterns: []   # default: built-in list, can extend with custom regex
```

### Note
This is input-level filtering only. For production use with untrusted input, consider running the tool in a separate subprocess or container with restricted permissions.

### Checklist
- [x] DCO signed
- [x] Unit tests pass (12/12)
- [x] Backward compatible (sandbox is opt-out)

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/python_repl.py
+++ b/agentuniverse/agent/action/tool/common_tool/python_repl.py
@@ -7,34 +7,82 @@
 # @FileName: python_repl.py
 
 import re
+from typing import List, Optional
 
 from agentuniverse.agent.action.tool.tool import Tool, ToolInput
 from langchain_community.utilities import PythonREPL
 from pydantic import Field
 
+_DEFAULT_BLOCKED_PATTERNS = [
+    r"__import__\s*\(",
+    r"\bimport\s+os\b",
+    r"\bfrom\s+os\b",
+    r"\bimport\s+subprocess\b",
+    r"\bfrom\s+subprocess\b",
+    r"\bimport\s+shutil\b",
+    r"\bfrom\s+shutil\b",
+    r"\bimport\s+socket\b",
+    r"\bfrom\s+socket\b",
+    r"\bimport\s+ctypes\b",
+    r"\bfrom\s+ctypes\b",
+    r"\bimport\s+sys\b",
+    r"\bfrom\s+sys\b",
+    r"\bopen\s*\([^)]*['\"]?[/~]",
+    r"os\.system\s*\(",
+    r"os\.popen\s*\(",
+    r"subprocess\.",
+    r"eval\s*\(",
+    r"exec\s*\(",
+    r"compile\s*\(",
+    r"globals\s*\(\s*\)",
+    r"locals\s*\(\s*\)",
+    r"getattr\s*\(",
+    r"setattr\s*\(",
+    r"delattr\s*\(",
+]
+
 
 class PythonREPLTool(Tool):
-    """The mock search tool.
+    """Python REPL tool that executes Python code.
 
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
The direction is valuable: making `PythonREPLTool` safer is aligned with the project, especially because agents may generate code automatically. I cannot merge this implementation yet because the blacklist gives a sandbox label while leaving straightforward bypasses open.

The current `_check_safety()` passes several dangerous inputs without executing them, for example:

- `import importlib; importlib.import_module("os").system("id")`
- `from pathlib import Path; Path("/etc/passwd").read_text()`
- `__builtins__.__dict__["__import__"]("os").system("id")`

All three returned `None` from `_check_safety()` locally, so they would be handed to the REPL. The unit tests pass, but they only cover the exact blocked strings, not the obvious alternate access paths.

Please either move this toward a safer implementation, such as an AST-based allowlist for a deliberately small subset of Python, or narrow the feature/wording so it is not presented as a sandbox. In either case, add regression tests for bypass-style inputs like the above. Security improvements are worth doing here; I just do not want us to merge a guardrail that looks stronger than it is.

hxbszyy:
The security goal has now been addressed by #608, which disables Python execution by default and requires an explicit opt-in without claiming to provide isolation. This blacklist remains bypassable through ordinary alternate import and file APIs, so keeping it open would encourage a weaker and misleading security model alongside the shipped behavior. I am closing this as superseded.

PR #582, opened 2026-06-15
fix(tool): restrict file tools to base directory

Loyal-Young:
**Checklist | 检查项**
- [x] I have read and understood the [contributor guidelines](https://github.com/antgroup/agentUniverse/blob/master/CONTRIBUTING.md). | 我已阅读并理解[贡献者指南](https://github.com/antgroup/agentUniverse/blob/master/CONTRIBUTING_zh.md)。
- [x] I have checked for any duplicate features related to this request and communicated with the project maintainers. | 我已检查没有与此请求重复的功能并与项目维护者进行了沟通。
- [x] I accept the suggestion of the maintainers to make changes to or close this PR. | 我接受此PR配合维护人员的建议进行修改或关闭。
- [x] I have submitted the test files and can provide screenshots of the test results (required for feature or bug fixes) | 我已经提交了测试文件并可提供测试结果截图(功能修改、BUG修复类PR必须提供，其他按需)
- [ ] I have added or modified the documentation related to this PR | 我已经添加或修改了本次pr对应的文档说明(非必要，根据实际PR内容按需添加)
- [ ] I have added examples and notes if needed | 我已经添加了使用案例代码与文档说明(非必要，根据实际PR内容按需添加)

**Please fill in the specific details of this PR:** | **请详细填写本次PR的内容:**

Fixes #571.

This PR restricts `WriteFileTool` and `ViewFileTool` to a configurable base directory to prevent arbitrary filesystem reads/writes.

Changes:
- Add shared safe path resolution helper.
- Add `base_dir` support for `WriteFileTool` and `ViewFileTool`.
- Reject absolute paths and path traversal attempts that escape the allowed directory.
- Add tests for allowed file access and rejected unsafe paths.

**Please provide the path of test files and submit screenshots or files of the test results(fill in as needed):** | **请填写测试文件路径并提供测试结果截图或文件(按需填写):**

Test files:
- `tests/agent/action/tool/common_tool/test_file_path_utils.py`
- `tests/agent/action/tool/common_tool/test_write_file_tool.py`
- `tests/agent/action/tool/common_tool/test_view_file_tool.py`

Local checks:
- `python -m py_compile agentuniverse/agent/action/tool/common_tool/file_path_utils.py agentuniverse/agent/action/tool/common_tool/write_file_tool.py agentuniverse/agent/action/tool/common_tool/view_file_tool.py`
- `git diff --check`

Full unit tests were not run locally because required dependencies such as `langchain` are missing in the current environment.

[...]

PR #607, opened 2026-07-10
fix(tool): confine WriteFileTool/ViewFileTool paths to optional base_dir (#571)

### LINKED ISSUE #571: WriteFileTool and ViewFileTool accept arbitrary filesystem paths with no sandboxing
## Bug Description

`WriteFileTool` and `ViewFileTool` in `agentuniverse/agent/action/tool/common_tool/` take their `file_path` parameter directly from LLM/agent output with no path validation. Both tools operate on the full host filesystem without restriction.

## Location

- `agentuniverse/agent/action/tool/common_tool/write_file_tool.py:34`
- `agentuniverse/agent/action/tool/common_tool/view_file_tool.py:28`

## Reproduction

**WriteFileTool:** instruct an agent to write to `/etc/cron.d/backdoor`, `~/.bashrc`, or `~/.ssh/authorized_keys`. The tool creates parent directories and writes content to the path with no check.

**ViewFileTool:** instruct an agent to read `/etc/shadow`, `.env`, or any credential file. The tool opens and returns the full contents.

In user-facing deployments, prompt injection via untrusted document/web content can trigger either tool without explicit user instruction.

## Impact

- `WriteFileTool`: attackers can overwrite system configuration files, crontabs, or SSH authorized keys, achieving persistent access on the host.
- `ViewFileTool`: attackers can exfiltrate API keys, database credentials, private keys, or any other readable file.

## Suggested Fix

Add a configurable `base_dir` parameter that defaults to a safe working directory. Reject any resolved path that escapes `base_dir`:

```python
import os

def _safe_path(self, file_path: str) -> str:
    base = os.path.realpath(self.base_dir)
    resolved = os.path.realpath(os.path.join(base, file_path))
    if not resolved.startswith(base + os.sep):
        raise ValueError(f"Path {file_path!r} escapes the allowed directory")
    return resolved
```

Document the restriction clearly so integrators know the expected scope.

---

123123213weqw:
## What & why

Fixes #571.

`WriteFileTool` and `ViewFileTool` take `file_path` straight from agent/LLM output with no validation and operate on the full host filesystem. In deployments that process untrusted content, a prompt injection can read/write sensitive files (e.g. `~/.ssh/authorized_keys`, `.env`, `/etc/shadow`).

## Change

Add an optional `base_dir` attribute to both tools. When configured, `file_path` (relative or absolute) is resolved with `os.path.realpath` and must stay within `base_dir`; any path that escapes it — an absolute path, a `..` traversal, or a symlink — is rejected with an error.

```python
tool = WriteFileTool(base_dir='/var/au_workspace')
```

or from the tool yaml:

```yaml
base_dir: /var/au_workspace
```

**Backward compatible:** when `base_dir` is unset (the default), behaviour is identical to today — the tool keeps operating on the full filesystem. So existing users are unaffected unless they opt into confinement.

## Tests

- Fixed the pre-existing unit tests for these tools: they called `execute(ToolInput(...))` against the `execute(**kwargs)` signature, so they were failing on `master`. Switched them to `execute(file_path=..., ...)`.
- Added coverage for the new behaviour: path inside `base_dir` allowed, relative path resolved under `base_dir`, absolute path outside rejected, `..` traversal rejected, read of an existing file outside `base_dir` rejected.

```
12 passed
```

## Notes

- This is an opt-in hardening (default unchanged), so it does not change existing behaviour for anyone who does not set `base_dir`.
- Happy to adjust naming/error format to match project conventions.


hxbszyy (changes requested):
The path-resolution code is reasonable and the 12 focused tests pass, but this does not actually close #571 because base_dir defaults to None. In the default configuration, both tools can still read or write any host path, which is the vulnerability the issue describes. An opt-in sandbox is useful as an escape hatch, not as the security default. Please make confinement the default using a clearly defined workspace, document how an integrator can override it, and add tests proving that the default rejects paths outside the workspace and blocks symlink escapes. This also overlaps #582, so we should merge one safe-by-default implementation rather than ship two competing variants.

123123213weqw:
Thanks for the careful review — you're right that the opt-in `base_dir` left the default configuration wide open, so it didn't actually close #571. I've reworked it so confinement is the default.

## What changed

**Confinement is now on by default.** When `base_dir` is unset (the default), the workspace resolves to the current working directory, so both `WriteFileTool` and `ViewFileTool` reject anything outside it out of the box. There is no longer an "unconfined by default" path.

The workspace is a clearly defined, single directory. The `_resolve_path` helper:
- picks `base_dir` when set, otherwise `os.getcwd()` as the default workspace;
- resolves the requested path with `os.path.realpath`, so `..` segments and symlinks cannot escape;
- raises `ValueError` (caught and returned as an error JSON payload) when the resolved path is not within the workspace.

**How an integrator overrides it.** Set `base_dir` on the tool, e.g. from the tool yaml:
```yaml
base_dir: /var/au_workspace
```
This widens/relocates the workspace to wherever the integrator wants file access scoped. The behavior is documented in the class and method docstrings (Google style).

## Tests proving the safe default

I added focused tests for the *default* configuration (`base_dir` left unset) on each tool:
- `test_default_workspace_rejects_path_outside` — with the tool at its default, a file under a separate directory (outside the current working directory) is rejected.
- `test_default_workspace_blocks_symlink_escape` — a symlink created *inside* the workspace that points to a file outside is rejected (and for the write tool the outside target is verified untouched).

The existing configured-workspace, absolute-path, and `..` traversal tests are retained and renamed for clarity. Functional read/write tests now run inside an explicit workspace.

All 16 tests in the two modules pass:
```
tests/test_agentuniverse/unit/agent/action/tool/test_view_file.py  (7)
tests/test_agentuniverse/unit/agent/action/tool/test_write_file.py (9)
```

## On #582

Agreed we should ship one safe-by-default implementation rather than two variants. This revision is intended to be that single variant — please feel free to close the competing #582 approach in favor of this one, or let me know if you'd prefer the config field named differently to match that discussion.

## Notes / residual risk

- Confinement is enforced at path-resolution time via `realpath`, which blocks symlink escapes at check time. As noted in the docstrings, for fully untrusted input a restricted subprocess/container is still the strongest defense (same guidance PythonREPLTool's sandbox carries).
- The default workspace being the current working directory is the minimally invasive safe default; I did not add a separate "disable confinement" toggle because that would reintroduce the vulnerability the issue is about. If you'd like an explicit opt-out knob for trusted integrators, I can add one (e.g. a `confine` flag) — happy to do so on request.

hxbszyy:
The safe-by-default revision is the right direction, but #582 is now merged and solves #571 with a shared path resolver instead of duplicating confinement logic in both tools. The latest force-push here also carries unrelated reversions of the merged PubMed and LangChain work in #604 and #610. With the security fix already landed, there is no remaining mergeable delta in this branch, so I am closing it as superseded.

PR #608, opened 2026-07-10
fix(tool): disable PythonREPLTool by default, require explicit opt-in (#570)

### LINKED ISSUE #570: PythonREPLTool executes arbitrary code without sandboxing

123123213weqw:
## What & why

Fixes #570.

`PythonREPLTool` passes its `input` straight to `langchain_community.utilities.PythonREPL.run()`, which calls `exec()` on the host with no sandboxing, subprocess isolation, or resource limits. In any agent pipeline exposed to user input, a prompt injection reaching this tool escalates directly to arbitrary code execution.

## Change

Add an `allow_code_execution` flag (default `False`). The tool now refuses to run code and returns a clear, instructive error until an integrator explicitly opts in:

```python
tool = PythonREPLTool(allow_code_execution=True)
```

or from the tool yaml:

```yaml
allow_code_execution: true
```

Also replaced the stale, copy-pasted docstring ("mock search tool / BYD / Warren Buffett") with an accurate description and a prominent security warning.

## ⚠️ Behaviour change (intentional)

This **is** a breaking change on purpose: the tool no longer executes code unless `allow_code_execution` is set to `True`. That is exactly the risk the issue asks to remove — executing arbitrary code by default is the bug. If the maintainers prefer a non-breaking variant (warn-by-default instead of refuse-by-default), I'm happy to switch the default; flagging it here for an explicit decision.

## Tests

New `test_python_repl.py`:

- `test_disabled_by_default` — without opt-in the tool refuses and the message tells the integrator how to enable it.
- `test_opt_in_execuates` — with `allow_code_execution=True` the code runs and prints the expected output.

```
2 passed
```

## Notes

- No other code in the repo references `PythonREPLTool`, so the blast radius of the default change is limited to users who instantiate this tool directly.
- I did not attempt a fake subprocess "sandbox" — isolating Python reliably needs OS-level privilege dropping that a library can't do portably, so the honest fix is an opt-in gate + a loud warning, matching option 2/3 in the issue.


hxbszyy (changes requested):
The direction is right: an honest opt-in gate is safer than pretending an in-process Python sandbox is secure, and the focused tests pass. The integration is incomplete, though. This changes the default to disabled while the shipped python_runner YAML files and the English/Chinese guides still say the tool executes code directly; none of them exposes or explains allow_code_execution. Existing demos that use the common PythonREPLTool will silently stop working. Please update the affected built-in YAMLs and docs to make the policy explicit, decide which examples intentionally opt in, and add a config-level test for that behavior. Then this will be ready for another look.

123123213weqw:
Addressed the review feedback and re-ran the verification. Thanks for the review.

hxbszyy (changes requested):
The opt-in behavior, shipped YAML flags, and warning text now address the earlier product-level feedback. The branch itself is not safe to merge, though: the latest commit also rewrites pubmed_tool.py, its tests/docs/YAML, and agent.py back toward snapshots from before #604 and #610. Those are unrelated, already-merged changes and would silently remove current PubMed capabilities and the canonical LangChain import. Please rebuild this branch from current master and carry only the PythonREPL code, configs, docs, and tests. Once the diff is clean, I can review the actual security change on its merits.

123123213weqw:

123123213weqw:
@hxbszyy Thanks — the branch is rebuilt from current master (`b142affb`) so the diff now contains **only** the PythonREPL opt-in change:

- `python_repl.py`, the 3 built-in `python_repl_tool.yaml` files, the en/zh integrated-tools docs (PythonRepl section only), and `test_python_repl.py`.
- Dropped the reversions of #604 (`pubmed_tool.py` / `pubmed_tool.yaml` / `test_pubmed_tool.py`) and #610 (`agent.py`), and the zh-doc hunk that only reverted #604's expanded PubMed section.

The opt-in behavior, YAML flags, and warning text are unchanged from the version you already approved. `pytest test_python_repl.py` → 5 passed.

123123213weqw:
Hi @hxbszyy, gentle nudge - the points from your last review are addressed in the latest push (see my comment above: rebuilt from current master with a clean diff, and the shipped YAMLs and en/zh docs now expose allow_code_execution; `pytest test_python_repl.py` -> 5 passed). Could you take another look when you have a moment? Thanks!

hxbszyy (approved):
This is in the right shape now. The default is genuinely safe, the shipped demos opt in explicitly instead of inheriting a risky default, and the docs are clear about the host-level execution boundary. The real YAML configuration path is covered as well. I reran the focused suite in the agentUniverse environment; it passes. Approving.

PR #609, opened 2026-07-10
feat(knowledge): ReciprocalRankFusionProcessor for multi-channel recall fusion (#248)

### LINKED ISSUE #248: 【Easy Issue/入门实战】Enhancements for the knowledge post-processing components/知识召回后置处理组件贡献
======================== English version ==============================
agentUniverse currently provides default knowledge recall post-processing component. We requires other processing component, and hope you can focus on the following directions:

Rerank: Reranking of recalls (you can refer to the Dashscope rerank component in aU).
Other post-processing component for Fusion, filtering, summarization.

You can optimize and submit the PR by combining them with specific practical application scenarios.

======================== 中文版 ==============================
agentUniverse目前提供默认的知识召回后置处理组件，项目需要其他的加工手段，期待您建设的方向如下：
* rerank：召回重排（可参考目前au接入的dashscope rerank组件）
* 融合、过滤、摘要、总结：

可结合具体场景，对于召回内容进行融合、过滤、摘要或总结，进一步提升多路召回后内容的质量

123123213weqw:
## What

Implements the **fusion** direction of #248 — a recall post-processor that merges the ranked results of multiple retrieval channels (vector / keyword / graph / …) into a single relevance ranking via **Reciprocal Rank Fusion**:

```
score(d) = Σ_channels  1 / (k + rank_channel(d))
```

RRF combines ranked lists without needing the (incomparable) raw scores of each channel. A document that appears near the top of several channels accumulates a higher fused score than one that only appears in a single channel — exactly the multi-channel recall-quality goal of #248. This complements the summarization processor (the summary direction of the same issue).

## Why this design

- **Channel-aware.** Channels are told apart by a metadata field (`channel_key`) that each retrieval channel stamps onto its docs. The processor scores each list independently and sums per-document.
- **Safe default.** With `channel_key` unset (or absent on the docs) the whole input is treated as one channel and the processor degrades to deterministic position-based scoring that **preserves input order** — so it can be dropped into an existing single-store pipeline without changing behavior.
- **De-duplication.** The same document surfacing in several channels is counted once per channel and merged by identity (deterministic `id` by default; configurable to `text` or a metadata field).
- **Non-destructive.** The fused score is written into `metadata[score_field]`; document `id` / `embedding` / `keywords` / `text` are preserved.

## Config

| Field | Default | Purpose |
| --- | --- | --- |
| `channel_key` | `''` | metadata field identifying a doc's retrieval channel |
| `dedup_key` | `'id'` | cross-channel identity (`id` / `text` / metadata field) |
| `k` | `60` | RRF smoothing constant (conventional value) |
| `top_n` | `null` | cap on returned docs |
| `score_field` | `'relevance_score'` | metadata key for the fused score |

## Example

```python
proc = ReciprocalRankFusionProcessor(channel_key="channel", k=60)
# vector channel:  A(0), B(1), C(2)
# keyword channel: A(0), D(1), E(2)
proc.process_docs([...])
# -> A = 2/61 (top of both channels), then B/D each 1/62, ...
```

## Tests

15 unit tests added under `tests/test_agentuniverse/unit/agent/action/knowledge/doc_processor/test_reciprocal_rank_fusion_processor.py`, all green locally:

- single-channel fallback (order preserved, scores `1/(k+r+1)`, metadata preserved)
- multi-channel fusion + exact RRF formula
- cross-channel and intra-channel dedup (by `id` and by `text`)
- channel-key-absent fallback
- `top_n` bounding (incl. `0`), custom `score_field`
- `_initialize_by_component_configer` loading + defaults

```
======================== 15 passed, 3 warnings in 0.22s ========================
```

(The warnings are pre-existing pydantic-v2 deprecation notices in the framework, unrelated to this change.)

Closes #248 (fusion direction).

### DIFF
--- a/agentuniverse/agent/action/knowledge/doc_processor/reciprocal_rank_fusion_processor.py
+++ b/agentuniverse/agent/action/knowledge/doc_processor/reciprocal_rank_fusion_processor.py
@@ -1,0 +1,214 @@
+# !/usr/bin/env python3
+# -*- coding:utf-8 -*-
+
+# @Time    : 2026/07/10
+# @Author  : contributor
+# @FileName: reciprocal_rank_fusion_processor.py
+
+"""
+Reciprocal Rank Fusion (RRF) document processor.
+
+Recall post-processor that merges the results of several retrieval channels
+into a single relevance ranking, addressing the *融合 (fusion)* direction of
+issue #248 ("融合、过滤、摘要、总结...进一步提升多路召回后内容的质量").
+
+Reciprocal Rank Fusion is a parameter-light way to combine multiple ranked
+result lists without needing the (often incomparable) raw scores of each
+channel. For a document ``d`` the fused score is::
+
+    score(d) = Σ_channels  1 / (k + rank_channel(d))
+
+where ``rank_channel(d)`` is the 1-based position of ``d`` within the channel
+that returned it (``k`` is a smoothing constant, conventionally 60). A document
+that appears near the top of several channels therefore accumulates a higher
+fused score than one that only appears in a single channel, even if it ranks
+first there.
+
+The processor consumes a flat list of retrieved :class:`Document` objects.
+Channels are told apart by a metadata field (``channel_key``) that each
+retrieval channel is expected to stamp onto its documents.
+:meth:`Knowledge.query_knowledge` does this automatically: it stamps every
+recalled document with the store code under the ``recall_channel`` key and
+keeps per-store ranked results, so fusing multi-store recall works without
+extra wiring. When no document carries the channel field the whole input is
+treated as a single channel, in which case RRF degrades to deterministic
+position-based scoring that preserves the input order — making the component
+safe to drop into an existing single-store pipeline.
+"""
+
+from typing import Dict, List, Optional, Tuple
+
+from agentuniverse.agent.action.knowledge.doc_processor.doc_processor import DocProcessor
+from agentuniverse.agent.action.knowledge.store.document import Document
+from agentuniverse.agent.action.knowledge.store.query import Query
+from agentuniverse.base.config.component_configer.component_configer import \
+    ComponentConfiger
+
+
+class ReciprocalRankFusionProcessor(DocProcessor):
+    """Fuses the ranked results of multiple retrieval channels via RRF.
+
+    Attributes:
+        channel_key (Optional[str]): Metadata field used to tell channels
+            apart. Each retrieval channel should write its identifier into
+            ``document.metadata[channel_key]``. When ``None`` (or when no
+            document carries the field) the whole input is treated as one
+            channel.
+        dedup_key (str): Identity used to deduplicate the same document
+            appearing in several channels. ``"id"`` (the default) uses the
+            deterministic document id; ``"text"`` compares raw content; any
+            other value is read from ``document.metadata``. When the configured
+            metadata field is absent from a document, its unique id is used as
+            a safe fallback so such documents are not all collapsed together.
+        k (int): RRF smoothing constant. Larger values dampen the advantage of
+            top ranks; the conventional value is 60.
+        top_n (Optional[int]): Maximum number of fused documents to return.
+            ``None`` returns all.
+        score_field (str): Metadata key the fused score is written into on each
+            returned document.
+    """
+
+    channel_key: Optional[str] = None
+    dedup_key: str = "id"
+    k: int = 60
+    top_n: Optional[int] = None
+    score_field: str = "relevance_score"
+
+    def _process_docs(self, origin_docs: List[Document],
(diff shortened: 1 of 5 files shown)


hxbszyy (changes requested):
RRF is valuable and squarely aligned with #248, and the 15 processor tests pass. The blocker is the real Knowledge integration. Knowledge.query_knowledge currently merges all store futures into one dict keyed by document id before _rag_post_process runs, and it drops the store code along the way. By the time this processor sees the data, a document returned by two stores has already been collapsed to one copy and its channel identity is gone, so the cross-channel score accumulation demonstrated by the unit tests cannot happen in the normal pipeline. Please preserve ranked channel results or stamp the store/channel before global deduplication, then add an integration test through Knowledge with two stores where the same document appears in both and receives a combined RRF score. Also, when a custom dedup_key is missing from metadata, the current identity is None for every such document and collapses them all; that needs a safe fallback and a test.

123123213weqw:
Thanks for the thorough review — both points are addressed in the latest push.

## 1. Real Knowledge integration

`Knowledge.query_knowledge` no longer collapses the cross-store result before post-processing. It now:

- stamps each recalled document with its source store code under a `recall_channel` metadata key (`RECALL_CHANNEL_KEY`), and
- de-duplicates **per channel** (`(id, store)` instead of `id` alone),

so a document recalled by several stores is kept once per store and reaches the RRF processor with its channel identity intact. The processor's `channel_key` default is now `recall_channel`, so multi-store fusion works out of the box, and it still degrades to single-channel position scoring when no document carries the field — single-store pipelines are unaffected.

A new integration test (`test_knowledge_rrf_integration.py`) drives the real `query_knowledge` path with two stores that both recall the same document and asserts it receives the combined score `2 * 1/(k+1)` and ranks first, while the single-channel-only docs tie at `1/(k+2)`.

## 2. Custom dedup_key fallback

`_identity` now falls back to the document id when the configured metadata field is missing, so documents lacking the field are no longer all collapsed to a single `None` identity. New unit tests cover the missing-field case, mixed presence, and custom-field dedup.

## Verification

All 21 processor + integration tests pass locally; the full suite is left to CI.

hxbszyy (changes requested):
The new integration test proves that RRF can now see both channels, but it does so by changing Knowledge.query_knowledge for every pipeline: cross-store deduplication moves from document id to (document id, store). Any knowledge config without RRF will therefore start returning duplicate documents that master currently removes. An optional fusion processor cannot change the default retrieval contract for everyone else. Please introduce a channel-aware handoff that preserves today's output when RRF is not selected, and add a regression test for the ordinary multi-store/no-RRF path. This branch also contains unrelated reversions of #604 and #610 after the latest rewrite; rebuild it from current master before the next review.

123123213weqw:

123123213weqw:
@hxbszyy The branch is rebuilt from current master (`2f2dffb2`). The diff now contains **only** the RRF processor + its yaml + the `knowledge.py` change + tests. Dropped the reversions of #604 / #610 and the zh-doc hunk.

The default-contract concern is addressed — channel-aware recall is now **opt-in**:

- `Knowledge.query_knowledge` keeps per-channel copies of a document (stamps `recall_channel`, dedups by `(id, store)`) **only** when a configured post-processor declares a non-`None` `channel_key` — i.e. an actual fusion processor that needs to tell channels apart.
- With no such processor, the merge keeps the default retrieval contract exactly: a document recalled by several stores is returned **once** (first recall wins), keyed by `id`, and no `recall_channel` metadata is stamped. So an optional fusion processor can no longer change the output of pipelines that do not use it.

Added two regression tests through the real `query_knowledge` path:
- ordinary multi-store / **no-RRF** (`post_processors=[]`): shared doc collapsed to one copy, no channel metadata leaked;
- a processor with `channel_key=None`: default contract still preserved.

`pytest test_reciprocal_rank_fusion_processor.py test_knowledge_rrf_integration.py` → 23 passed.

123123213weqw:
Hi @hxbszyy, gentle nudge - your last review is addressed in the latest push (see my comment above: channel-aware recall is now opt-in, so the default retrieval contract is preserved when no fusion processor is configured, with regression tests for the no-RRF path; rebuilt from current master with a clean diff; `pytest test_reciprocal_rank_fusion_processor.py test_knowledge_rrf_integration.py` -> 23 passed). Could you take another look when you have a moment? Thanks!

hxbszyy (approved):
The rebuild addresses the compatibility concern cleanly. Existing pipelines keep the original cross-store dedup behavior, while per-channel copies are preserved only when a configured processor explicitly asks for channel metadata. That gives us useful multi-retriever fusion without quietly changing every knowledge query. The focused processor and integration tests pass in the agentUniverse environment. Approving.

Issue #252, opened 2025-01-16
【Easy Issue/入门实战】Enhancements for more tools/工具插件接入


LandJerry:
======================== English version ==============================
The expected tools are as follows：

* Search tools
* Code Interpreter tools
* Other tools
  * Youtube
  * Yahoo Finance
  * ArXiv
  * office
  * slash
  * others
 
======================== 中文版 ==============================
期待您优先接入的工具插件如下：

* Search类工具
* Code Interpreter类工具
* 其他工具
  * Youtube
  * Yahoo Finance
  * ArXiv
  * office
  * slash
  * others

wilkinszhang:
@LandJerry I  want to have a try, plz assign to me, thanks.

weizjajj:
There is already a related PR for the ArXiv tool https://github.com/antgroup/agentUniverse/pull/301. You can submit other tools outside of this PR, and you can refer to the storage directory of the tools in this PR. Place the relevant tools in the agentuniverse/agent/action/tool/common_tool directory. In the dev branch, under this directory, some tools already exist, which you can reference to create new tools.Additionally, you can refer to this test_case https://github.com/antgroup/agentUniverse/pull/301/files#diff-49144f14450a5ddd14dcf925e7ebfbef6f804ad0839a7d558b277226e0f58717 to create test cases for the relevant tools.

xmhu2001:
@LandJerry Hi, I want to build a YouTube tool that can search videos by keyword, retrieve channel details, and fetch trending videos. Could you assign this issue to me?

Veteran-ChengQin:
@LandJerry Hello, I would like to contribute to the **Code Interpreter tools**. However, the term itself is somewhat unclear in terms of specific requirements. Could you elaborate on this?

Metastarx:
Hi, I would like to work on a PubMed Search Tool under this issue. Could you please assign it to me?

happym51888:
[...]

PR #610, opened 2026-07-10
fix(agent): use canonical langchain.tools import in agent.py (#190)

123123213weqw:
## What

Fixes #190. Replaces the two deprecated `from langchain.agents.tools import Tool` imports in `agentuniverse/agent/agent.py` with the canonical `from langchain.tools import Tool as LangchainTool`.

## Why

- **`langchain.agents.tools` is the legacy re-export location.** It still resolves on the currently pinned langchain, but it has been removed in newer langchain releases. Importing from `langchain.tools` is the supported path.
- **Consistency.** The rest of the codebase already imports the langchain Tool from the canonical path under the `LangchainTool` alias:

  ```
  agentuniverse/agent/plan/planner/react_planner/react_planner.py
  agentuniverse/agent/plan/planner/nl2api_planner/nl2api_planner.py
  agentuniverse/agent/template/react_agent_template.py
  agentuniverse/agent/action/knowledge/knowledge.py
  agentuniverse/agent/action/tool/tool.py
  ```

  `agent.py` was the only outlier.

- **Why the alias & local import.** `agent.py` already imports the agentUniverse `Tool` at module top (`from agentuniverse.agent.action.tool.tool import Tool`). Keeping the langchain import local to the two methods that use it — and aliasing it as `LangchainTool` — avoids shadowing that class, matching the convention used elsewhere.

## Changes

`as_langchain_tool` and `async_as_langchain_tool`:

```diff
-        from langchain.agents.tools import Tool
+        from langchain.tools import Tool as LangchainTool
         ...
-        return Tool(
+        return LangchainTool(
```

No behavior change — `langchain.tools.Tool` and `langchain.agents.tools.Tool` resolve to the same class (`langchain_core.tools.Tool`) on supported langchain versions.

## Verification

- `python -m py_compile agentuniverse/agent/agent.py` — clean.
- Confirmed the canonical import resolves: `from langchain.tools import Tool` → `<class 'langchain_core.tools.Tool'>`.
- No remaining `from langchain.agents.tools import` references in the repo.

Closes #190.

PR #611, opened 2026-07-10
feat(tool): add YahooFinanceTool for market data (#252)

123123213weqw:
## What

Adds the **Yahoo Finance** tool requested in #252 — a market-data tool that fetches the latest quote, historical OHLCV bars, or the company profile for a stock / ETF / index symbol (e.g. `AAPL`, `600519.SS`).

```python
tool = YahooFinanceTool()
tool.execute(mode="quote",   symbol="AAPL")
tool.execute(mode="history", symbol="AAPL", period="3mo", interval="1d")
tool.execute(mode="info",    symbol="TSLA")
```

## Design

- **Three modes** via a single `execute(mode, symbol, period, interval)` entry point:
  - `quote` — latest price, change, % change, volume, market cap, 52w range…
  - `history` — OHLCV bars for a configurable `period` / `interval`, newest-first, capped at `max_history_rows` to keep the agent context compact.
  - `info` — full company profile (name, sector, industry, summary…).
- **Optional dependency, lazily loaded.** Built on `yfinance` (the de-facto Python client for Yahoo Finance data), but `yfinance` is imported inside the methods that need it — so importing the module never requires the dependency, and a missing `yfinance` surfaces a clear `pip install yfinance` hint only when the tool is actually invoked. An agent that never calls the tool pays no install cost.
- **Network isolated from formatting.** The `_fetch_*` helpers (which hit yfinance) are separate from the pure `_format_*` helpers. This makes the component fully unit-testable without a network connection.
- **Robustness.** `@retry` on every fetch, symbols normalized (uppercased/trimmed), empty/invalid inputs and fetch failures return human-readable error strings instead of raising.

## Config

Registered as `yahoo_finance_tool` with `args_model` + `args_model_schema` (mirroring `github_tool`), and exported from `agentuniverse.agent.action.tool.common_tool`.

## Tests

17 deterministic unit tests added under `tests/test_agentuniverse/unit/agent/action/tool/test_yahoo_finance_tool.py`, all green locally. The yfinance layer is mocked, so the suite runs with **no network and no `yfinance` install**:

- pure formatters: quote / history / info (with data, empty, missing fields)
- numeric helpers: compact market-cap suffixes (T/B/M/K), symbol normalization
- `execute` wiring: quote / info (mock ticker), history (end-to-end via a tiny DataFrame stub, incl. newest-first ordering)
- symbol uppercasing before fetch
- invalid mode, missing symbol, missing-`yfinance` install hint, fetch-error surfacing

```
======================== 17 passed, 3 warnings in 0.26s ========================
```

(The warnings are pre-existing pydantic-v2 deprecation notices in the framework, unrelated to this change.)

Partially addresses #252 (Yahoo Finance item).

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/__init__.py
+++ b/agentuniverse/agent/action/tool/common_tool/__init__.py
@@ -7,7 +7,9 @@
 # @FileName: __init__.py
 
 from .github_tool import GitHubTool
+from .yahoo_finance_tool import YahooFinanceTool
 
 __all__ = [
-    'GitHubTool'
+    'GitHubTool',
+    'YahooFinanceTool',
 ]
--- a/agentuniverse/agent/action/tool/common_tool/yahoo_finance_tool.py
+++ b/agentuniverse/agent/action/tool/common_tool/yahoo_finance_tool.py
@@ -1,0 +1,299 @@
+# !/usr/bin/env python3
+# -*- coding:utf-8 -*-
+
+# @Time    : 2026/07/10
+# @Author  : contributor
+# @FileName: yahoo_finance_tool.py
+
+"""
+Yahoo Finance market-data tool.
+
+Provides stock / ETF / index quotes, historical OHLCV bars, and company
+information sourced from Yahoo Finance, addressing the *Yahoo Finance* item of
+issue #252 (tool plug-in integration).
+
+The tool is built on top of the ``yfinance`` library (the de-facto Python
+client for Yahoo Finance data). ``yfinance`` is imported lazily inside the
+methods that need it, so importing this module never requires the dependency —
+an agent that never calls the tool pays no install cost, and a missing
+``yfinance`` surfaces a clear, actionable error only when the tool is actually
+used.
+
+Three operating modes are supported:
+
+* ``quote``   — the latest price, change, and a handful of headline metrics.
+* ``history`` — historical OHLCV bars for a configurable period / interval.
+* ``info``    — a curated company profile (a stable subset of the Yahoo
+  Finance profile; the full ``ticker.info`` payload is intentionally not
+  dumped, to keep the agent context compact).
+
+All output is a human-readable string so it can be dropped directly into an
+agent's context. The data-fetching helpers are isolated from the formatting
+helpers, which keeps the component fully unit-testable without a network
+connection: tests stub :meth:`_get_ticker` and feed fixed dicts to the pure
+formatters.
+"""
+
+from typing import Any, Dict, List, Optional
+
+from pydantic import Field
+
+from agentuniverse.agent.action.tool.tool import Tool
+from agentuniverse.base.annotation.retry import retry
+
+
+class YahooFinanceTool(Tool):
+    """Yahoo Finance market-data tool.
+
+    Attributes:
+        max_history_rows (int): Upper bound on the number of history rows
+            returned, to keep the agent context compact.
+        quote_fields (List[str]): Keys surfaced in ``quote`` mode, in display
+            order.
+        info_fields (List[str]): Curated keys surfaced in ``info`` mode, in
+            display order. Only these keys are emitted, so the agent context
+            is not flooded by the large, unstable ``ticker.info`` payload.
+    """
+
+    name: str = "yahoo_finance_tool"
+    description: str = (
+        "Yahoo Finance market-data tool. Fetches the latest quote, historical "
+        "OHLCV bars, or the company profile for a stock / ETF / index symbol "
+        "(e.g. 'AAPL', '600519.SS')."
+    )
+    max_history_rows: int = Field(30, description="Max number of history rows returned.")
+    quote_fields: List[str] = [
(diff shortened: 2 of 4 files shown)


hxbszyy (changes requested):
Yahoo Finance is explicitly useful for #252, and the implementation has a solid deterministic test base: all 17 focused tests pass. The shipped component is not registerable yet, though. Loading yahoo_finance_tool.yaml through the framework returns component_type=None because class_path is not part of the component metadata contract; the scanner expects metadata.type/module/class or meta_class. The current tests instantiate YahooFinanceTool directly, so they miss this failure. Please switch the YAML to the supported component schema and add a config/registration test that resolves and initializes the tool through the real loader. I would also keep info-mode output curated rather than appending every ticker.info field, since that can flood an agent context with a very large and unstable payload.

123123213weqw:
Thanks for the thorough review — all three points are addressed in this update.

## 1. Registration (`component_type=None`)

**Root cause:** `yahoo_finance_tool.yaml` declared the class via `class_path`, which is not part of the component metadata contract. `ComponentConfiger.get_component_config_type()` therefore returned `None`, so the scanner never matched it against the `TOOL` type and silently skipped registration — which is why loading the YAML produced a non-registerable component.

**Fix:** the YAML now uses the supported schema, consistent with the other builtin tools (pubmed / arxiv):

```yaml
tool_type: api
metadata:
  type: TOOL
  module: agentuniverse.agent.action.tool.common_tool.yahoo_finance_tool
  class: YahooFinanceTool
```

Verified through the real loader: `Configer → ComponentConfiger → get_component_config_type()` now returns `TOOL` (it was `None` with the old `class_path`).

## 2. Config / registration test through the real loader

Added a `TestRegistration` class that loads the **shipped** `yahoo_finance_tool.yaml` and runs it through the actual pipeline the component scanner uses:

- `ComponentConfiger.get_component_config_type() == TOOL` — the exact gate that returned `None` before;
- `ToolManager().get_instance_obj("yahoo_finance_tool")` resolves, initializes, and returns a `YahooFinanceTool` with the expected `name`, `input_keys`, and `args_model_schema`;
- the resolved instance actually `run()`s (network stubbed).

These cases fail on the old `class_path` schema and pass on the new one, so the gap the existing tests missed is now covered.

## 3. Curated info-mode output

`info` mode now emits only a curated `info_fields` allowlist (identity, exchange/sector/industry, valuation, margins, balance-sheet headline items, 52-week range, etc.), skipping empty values — instead of appending every `ticker.info` key. Unknown / unstable keys no longer leak into the agent context.

## Verification

All 23 focused tests pass:

```
pytest tests/test_agentuniverse/unit/agent/action/tool/test_yahoo_finance_tool.py
# 23 passed
```

The 6 new tests cover info curation (2) and registration (4). The suite remains fully deterministic — no network connection or `yfinance` install is required.

hxbszyy (changes requested):
The registration path and curated field list are much better, and this is the Yahoo Finance candidate I would keep working on. It still cannot merge from the current branch. The latest commit also modifies PubMed code/tests/docs and agent.py, effectively carrying reversions of #604 and #610 alongside the new tool. Please rebuild from current master so the diff contains only YahooFinanceTool, its config, exports, and tests. While doing that, cap longBusinessSummary (it is still an unbounded payload despite the curated list) and restore the ApplicationConfigManager state after registration tests so this suite cannot leak global configuration into tests that run after it.

123123213weqw:

123123213weqw:
@hxbszyy The branch is rebuilt from current master (`6202fdb9`). The diff now contains **only** `YahooFinanceTool` + its yaml + the `__init__` export + tests. Dropped the reversions of #604 / #610 and the zh-doc hunk (which only reverted #604's PubMed section).

Both points from your second review are addressed:

1. **`longBusinessSummary` is now bounded.** New `max_business_summary_chars` field (default 400); in `info` mode the field is truncated with an ellipsis, so even inside the curated `info_fields` list it can no longer dump an unbounded description. Configurable per-instance.

2. **Registration tests no longer leak global state.** `TestRegistration` snapshots `ApplicationConfigManager().app_configer` in `setUp` and restores it in `tearDown` (handling the unset case), so the synthetic `AppConfiger` they install cannot leak into later tests.

`pytest test_yahoo_finance_tool.py` → 25 passed (added 2 tests for the summary cap, default + custom).

123123213weqw:
Hi @hxbszyy, gentle nudge - the review feedback is addressed in the latest push (see my comment above: the YAML now uses the supported component schema so the tool registers through the real loader, and info-mode output is curated; `pytest test_yahoo_finance_tool.py` -> 23 passed). Could you take another look when you have a moment? Thanks!

hxbszyy (approved):
This version now earns its place in the built-in tool set. It follows the real component registration path, keeps the optional dependency lazy, bounds both history and free-text output, and restores global config state in the tests. Among the overlapping Yahoo Finance proposals, this is the implementation I want us to keep. The focused suite passes in the agentUniverse environment. Approving.

PR #616, opened 2026-07-11
feat: add agent concurrency control and rate limiting

toni20050606:
- Add ThreadRateLimiter and AsyncRateLimiter based on semaphores
- Add AgentConcurrencyManager for global/single-agent concurrency control
- Add max_concurrent, concurrency_enabled, concurrency_timeout config to Agent
- Add unit tests and usage example

**Checklist | 检查项**
- [x] I have read and understood the [contributor guidelines](https://github.com/antgroup/agentUniverse/blob/master/CONTRIBUTING.md). | 我已阅读并理解[贡献者指南](https://github.com/antgroup/agentUniverse/blob/master/CONTRIBUTING_zh.md) 。
- [x] I have submitted the test files and can provide screenshots of the test results (required for feature or bug fixes) | 我已经提交了测试文件并可提供测试结果截图(功能修改、BUG修复类PR必须提供，其他按需)

-## What
Add concurrency control and rate limiting for agent execution.

## Why
- Prevent single agent from being overwhelmed by too many concurrent requests
- Support global and per-agent concurrency limits
- Provide async and sync rate limiter APIs

## How
- `ThreadRateLimiter` / `AsyncRateLimiter`: semaphore-based limiters
- `AgentConcurrencyManager`: singleton manager for global/per-agent control
- Agent base class supports `max_concurrent`, `concurrency_enabled`, `concurrency_timeout`
- Include unit tests and example

## Test
- Added tests/test_agentuniverse/unit/test_rate_limiter.py
- Added examples/concurrency_agent_example/
-
-

-

-

### DIFF
--- a/agentuniverse/agent/agent.py
+++ b/agentuniverse/agent/agent.py
@@ -5,6 +5,7 @@
 # @Author  : heji
 # @Email   : lc299034@antgroup.com
 # @FileName: agent.py
+import asyncio
 import json
 import uuid
 from abc import abstractmethod, ABC
@@ -13,6 +14,7 @@ from threading import Thread
 from typing import Optional, Any, List
 
 from langchain_core.runnables import RunnableSerializable, RunnableConfig
+
 from langchain_core.utils.json import parse_json_markdown
 
 from agentuniverse.agent.action.knowledge.knowledge import Knowledge
@@ -46,6 +48,7 @@ from agentuniverse.base.util.common_util import stream_output
 from agentuniverse.base.util.logging.logging_util import LOGGER
 from agentuniverse.base.util.memory_util import generate_messages, get_memory_string
 from agentuniverse.base.util.system_util import process_dict_with_funcs, is_system_builtin
+from agentuniverse.base.util.rate_limiter.rate_limiter import ThreadRateLimiter, AsyncRateLimiter
 from agentuniverse.base.tracing.au_trace_manager import AuTraceManager
 from agentuniverse.llm.llm import LLM
 from agentuniverse.llm.llm_manager import LLMManager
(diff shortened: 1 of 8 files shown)


hxbszyy (changes requested):
I cannot approve this concurrency model as implemented. _execution_contexts is a mutable class attribute, so every Agent instance shares the same counter; the len/check/insert sequence has no lock and can exceed the limit under the exact concurrency it is meant to control. async_run uses that same synchronous dictionary, concurrency_timeout is never used, and the new ThreadRateLimiter/AsyncRateLimiter/AgentConcurrencyManager classes are not wired into run or async_run at all. The PR also contains unrelated regressions: it replaces the canonical LangChain import fixed in #610 and changes process_prompt to pop inputs, including pop('audio_url') without a default, which can raise KeyError. Please reduce this to one design using the limiter primitives end to end, make state instance-scoped and thread/async safe, prove timeout behavior, and remove every unrelated Agent change.

hxbszyy (changes requested):
Thanks for the follow-up. This is closer, but the core async lifecycle is still broken, so I cannot approve it yet.

- `async_run()` acquires from `_async_limiter`, then its `finally` block calls `_release_concurrency_permit()`, which releases `_thread_limiter`. The async permit is never returned. After `max_concurrent` successful async calls, that agent remains exhausted, while the sync semaphore receives permits it never acquired. Please add and await an async release path, and report async activity correctly.
- The new agent-level tests do not currently validate the feature. On this head, the two focused tests fail in the `agentUniverse` environment: `SimpleAgent` never supplies a usable execution path and tracing is not isolated, so they fail before proving concurrency behavior. The barrier test can also wait forever because the worker threads never enter the barrier. We need real overlapping sync and async tests that fail before the fix and verify acquire, timeout/rejection, exception cleanup, and subsequent reuse.
- The example YAML never sets `concurrency_enabled: true`, so the configuration it advertises leaves the feature disabled.
- Please remove the unrelated `LangchainTool` to `StructuredTool` adapter rewrite from this PR. That is a separate public behavior change and should not ride along with concurrency control.

Once those are fixed, I will rerun the lifecycle tests and make a final call.

PR #620, opened 2026-07-12
fix(tool): lazy load YouTube API dependency

Loyal-Young:
**Checklist | 检查项**
- [x] I have checked for duplicate work.
- [x] I accept maintainer suggestions to modify or close this PR.
- [x] I have submitted the test files and test results.
- [ ] Documentation change is not needed for this fix.
- [ ] Example change is not needed for this fix.

**Please fill in the specific details of this PR:**

This PR makes `YouTubeTool` import optional YouTube API dependencies lazily.

Previously, importing `youtube_tool.py` required `googleapiclient` immediately. In environments without `google-api-python-client`, the module failed during import or test collection even if the tool was not used. This PR moves the dependency import into service initialization and returns a clear dependency error only when the tool is actually used.

Changes:
- Lazy-load `googleapiclient.discovery.build` in `_initialize_service()`.
- Provide a fallback `HttpError` class when `googleapiclient` is not installed.
- Update tests to use a fake YouTube service instead of real API calls.
- Keep search, channel info, and trending video behavior covered.

**Please provide the path of test files and submit screenshots or files of the test results:**

Test file:
- `tests/test_agentuniverse/unit/agent/action/tool/test_youtube_tool.py`

Checks:
```text
python -m pytest tests/test_agentuniverse/unit/agent/action/tool/test_youtube_tool.py -q
python -m py_compile agentuniverse/agent/action/tool/common_tool/youtube_tool.py tests/test_agentuniverse/unit/agent/action/tool/test_youtube_tool.py
git diff --check

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/youtube_tool.py
+++ b/agentuniverse/agent/action/tool/common_tool/youtube_tool.py
@@ -11,9 +11,13 @@ from pydantic import Field
 from agentuniverse.agent.action.tool.tool import Tool
 from agentuniverse.base.annotation.retry import retry
 from agentuniverse.base.util.env_util import get_from_env
-from googleapiclient.discovery import build
-from googleapiclient.errors import HttpError
 import re
+
+try:
+    from googleapiclient.errors import HttpError
+except ImportError:
+    class HttpError(Exception):
+        """Fallback used when google-api-python-client is not installed."""
 
 service_name = "youtube"
 api_version = "v3"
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
This fixes the right boundary. YouTube support is optional, so importing the rest of the tool package and collecting tests should not require google-api-python-client. Deferring service construction keeps that dependency at the point of use, and replacing live API tests with a fake service makes the coverage deterministic. The complete tool unit directory passes with this in the merge stack.

PR #621, opened 2026-07-12
fix(tool): guard readimage optional dependencies

Loyal-Young:
**Checklist | 检查项**
- [x] I have checked for duplicate work.
- [x] I accept maintainer suggestions to modify or close this PR.
- [x] I have submitted the test files and test results.
- [ ] Documentation change is not needed for this fix.
- [ ] Example change is not needed for this fix.

**Please fill in the specific details of this PR:**

This PR makes `readimage_tool` handle optional OCR dependencies more safely.

Previously, importing `readimage_tool.py` required packages such as `cv2`, `pytesseract`, `PIL`, and `numpy` at module import time. If any optional dependency was missing, test collection or module import failed immediately. This PR guards those imports and raises a clear dependency error only when image/OCR functionality is actually used.

Changes:
- Guard optional imports for `cv2`, `pytesseract`, `PIL`, and `numpy`.
- Add a shared dependency check helper with clear error messages.
- Keep non-OCR utility functions importable without image dependencies.
- Update tests to use the correct package import path.
- Skip only dependency-required tests when optional packages are missing.

**Please provide the path of test files and submit screenshots or files of the test results:**

Test file:
- `tests/test_agentuniverse/unit/agent/action/tool/test_readimage_tool.py`

Checks:
```text
python -m pytest tests/test_agentuniverse/unit/agent/action/tool/test_readimage_tool.py -q
python -m py_compile agentuniverse/agent/action/tool/common_tool/readimage_tool.py tests/test_agentuniverse/unit/agent/action/tool/test_readimage_tool.py
git diff --check

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/readimage_tool.py
+++ b/agentuniverse/agent/action/tool/common_tool/readimage_tool.py
@@ -6,18 +6,44 @@
 # @Email   : zhangdongxu0852@163.com
 # @FileName: readimage_tool.py
 
-import cv2
-import pytesseract
-from PIL import Image
-import numpy as np
 import re
 import os
+
+try:
+    import cv2
+except ImportError:
+    cv2 = None
+
+try:
+    import pytesseract
+except ImportError:
+    pytesseract = None
+
+try:
+    from PIL import Image
+except ImportError:
+    Image = None
+
+try:
+    import numpy as np
+except ImportError:
+    np = None
+
+
+def _require_dependency(dependency, package_name: str):
+    if dependency is None:
+        raise ImportError(
+            f"{package_name} is required for readimage_tool. "
+            f"Install it before using image OCR features."
+        )
+
 
 def enhance_image(image):
     """
     Enhance the image: convert to grayscale, apply CLAHE for contrast enhancement,
     then apply bilateral filtering to reduce noise while preserving edges.
     """
+    _require_dependency(cv2, "opencv-python")
     gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
     clahe = cv2.createCLAHE(clipLimit=2.0, tileGridSize=(8, 8))
     enhanced = clahe.apply(gray)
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
Guarding the Python imports is useful, but the revised tests are still environment-dependent. HAS_OCR_DEPS only checks importable packages; it does not check the tesseract executable or the requested language data. test_detect_text_regions_no_text calls the default frozen_east_text_detection.pb even though that model is not in the repository, so installing OpenCV turns the current skip into a deterministic failure. Please mock the EAST network and pytesseract boundary (or explicitly provide/check those assets), and add a test that imports the module with each optional package absent and asserts the clear runtime error. Right now the PR moves the failure rather than fully fixing the optional-dependency contract.

Loyal-Young:
> Guarding the Python imports is useful, but the revised tests are still environment-dependent. HAS_OCR_DEPS only checks importable packages; it does not check the tesseract executable or the requested language data. test_detect_text_regions_no_text calls the default frozen_east_text_detection.pb even though that model is not in the repository, so installing OpenCV turns the current skip into a deterministic failure. Please mock the EAST network and pytesseract boundary (or explicitly provide/check those assets), and add a test that imports the module with each optional package absent and asserts the clear runtime error. Right now the PR moves the failure rather than fully fixing the optional-dependency contract.

Updated the tests to avoid environment-dependent OCR/OpenCV assets.

The revised tests now mock the EAST DNN boundary and pytesseract.image_to_string, so they no longer require frozen_east_text_detection.pb, a tesseract executable, or language data. I also added coverage that imports the module with each optional dependency absent and verifies that the module import still succeeds while the relevant runtime call raises a clear ImportError.

hxbszyy (approved):
This fixes a real packaging problem without hiding runtime requirements. Importing the module no longer pulls in the full OCR stack, while each feature path still fails with a clear dependency message when it is actually used. The revised tests are deterministic and no longer depend on a local EAST model or Tesseract installation. The focused suite passes in the agentUniverse environment. Approving.

Issue #207, opened 2024-12-02
知识库我已经更新好了，为什么报错 Exception occurred in knowledge query: 'NoneType' object has no attribute 'query'


mymoonsoft:
我需要从知识库中查找关于预计负债核算错误的相关案例。这将帮助我理解常见的错误类型及其原因，从而更好地回答用户的问题。
Action: account_rule_knowledge
Action Input: 
```json
{
  "query_str": "预计负债 核算 错误 案例",
  "similarity_top_k": 5
}
```
Building prefix dict from the default dictionary ...
Loading model from cache C:\Users\Administrator\AppData\Local\Temp\jieba.cache
Loading model cost 0.352 seconds.
Prefix dict has been built successfully.
This is Query Result:
2024-12-02 18:29:46.384 | ERROR    | [default] | agentuniverse.agent.action.knowledge.knowledge:query_knowledge:190 | Exception occurred in knowledge query: 'NoneType' object has no attribute 'query'
2024-12-02 18:29:46.384 | ERROR    | [default] | agentuniverse.agent.action.knowledge.knowledge:query_knowledge:190 | Exception occurred in knowledge query: 'NoneType' object does not support the context manager protocol

mymoonsoft:
我还是copy回原来的版本，替换了数据库，数据库名称还是用原名称，内容替换,&nbsp;这样就可以用了


不过，这个对小白用户还是有些难度！！


------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "antgroup/agentUniverse"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2024年12月3日(星期二) 下午3:17
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [antgroup/agentUniverse] 知识库我已经更新好了，为什么报错 Exception occurred in knowledge query: 'NoneType' object has no attribute 'query' (Issue #207)





 
如果不是启动的web server可以检查下AgentUniverse.start()里面有没有带上core_mode=True的参数。
 可以单独确定一下store对应的数据库是否有对应的collection和表，保证知识插入成功了。

123123213weqw:
Took a look at this. Root cause is that `ComponentManagerBase.get_instance_obj` returns `None` silently when a store is not registered, so the later `.query` / `.insert_document` access throws the cryptic `'NoneType' object has no attribute 'query'` instead of telling you which store is missing.

[...]

PR #624, opened 2026-07-13
fix(knowledge): raise clear error for unregistered store/router (#207)

### LINKED ISSUE #203: law_knowledge.insert_knowledge(     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ AttributeError: 'NoneType' object has no attribute 'insert_knowledge'

123123213weqw:
## Problem

When a store (or rag-router) referenced by a Knowledge is not registered, the call site crashed with a cryptic, low-level error (#207): `'NoneType' object has no attribute 'query'` / `'NoneType' object has no attribute 'insert_knowledge'`.

> Note on #203: that report has a different root cause — the caller requested `account_rule_knowledge` while the YAML registered `account_case_knowledge`, i.e. a name mismatch that fails *before* a `Knowledge` instance is resolved. This PR's `strict` lookup only fires inside an already-resolved `Knowledge`, so it does **not** claim to fix #203; it targets the `NoneType`-on-missing-component crash in #207.

## Root cause

`ComponentManagerBase.get_instance_obj` returns `self._instance_obj_map.get(instance_code)` — i.e. **silently `None`** when the component is not registered. Every caller then does `…get_instance_obj(code).query` / `.insert_document`, so the `None` surfaces as an `AttributeError` deep inside the query/insert path, with no hint that a component is simply missing.

## Fix

1. **`component_manager_base.py`** — add an opt-in `strict: bool = False` flag to `get_instance_obj`. When `True` and the component is missing, it raises a descriptive `ValueError` naming the **component code** and **type** (e.g. `STORE component 'my_store' is not registered (resolved instance code '…'). Ensure the corresponding configuration is loaded before use — verify the component config file and the application bootstrap order.`) instead of returning `None`. **Default behaviour is unchanged** (`strict=False` still returns `None`).

2. **`knowledge.py`** — switch the four lookups to `strict=True`: the `query_knowledge` / `insert_knowledge` / `update_knowledge` store lookups and the `_route_rag` rag-router lookup.

## Rebased onto #609's channel-aware recall

The branch is rebuilt from current master, which now carries the channel-aware recall from #609. The `strict=True` change is preserved alongside the `(future, store_code)` pairs, the `recall_channel` metadata, and the opt-in deduplication contract — strict lookup and RRF fusion compose without either changing the other's behaviour.

## Tests

`tests/test_agentuniverse/unit/agent/action/knowledge/test_knowledge_registration_errors.py` (6):
- `TestGetInstanceObjStrict` (4): default still returns `None`; `strict` raises `ValueError` naming the code + type; happy path still returns a copy / raw instance under `strict`.
- `TestKnowledgeRegistrationErrors` (1): end-to-end through the real `query_knowledge` — an unregistered store raises a `ValueError` mentioning the store code and `STORE`, not an `AttributeError`.
- `TestStrictLookupWithChannelFusion` (1): with an RRF fusion processor configured, `query_knowledge` still resolves stores with `strict=True` and the per-channel merge fuses correctly — strict and channel-aware fusion coexist.

`python -m unittest test_knowledge_registration_errors` → **6 passed**, and the existing RRF integration tests (`test_knowledge_rrf_integration`) still pass.

Relates to #207. Part of #253 (error message optimization).

### DIFF
--- a/agentuniverse/agent/action/knowledge/knowledge.py
+++ b/agentuniverse/agent/action/knowledge/knowledge.py
@@ -187,7 +187,7 @@ class Knowledge(ComponentBase):
         for _store_code in stores:
             futures.append(
                 self.insert_executor.submit(
-                    StoreManager().get_instance_obj(_store_code).insert_document,
+                    StoreManager().get_instance_obj(_store_code, strict=True).insert_document,
                     document_list))
         wait(futures, return_when=ALL_COMPLETED)
         for future in futures:
@@ -213,7 +213,7 @@ class Knowledge(ComponentBase):
         for _store_code in stores:
             futures.append(
                 self.insert_executor.submit(
-                    StoreManager().get_instance_obj(_store_code).update_document,
+                    StoreManager().get_instance_obj(_store_code, strict=True).update_document,
                     document_list))
         wait(futures, return_when=ALL_COMPLETED)
         for future in futures:
@@ -225,7 +225,7 @@ class Knowledge(ComponentBase):
         LOGGER.info("Knowledge update complete.")
 
     def _route_rag(self, query: Query):
-        return RagRouterManager().get_instance_obj(self.rag_router).rag_route(query, self.stores)
+        return RagRouterManager().get_instance_obj(self.rag_router, strict=True).rag_route(query, self.stores)
 
     @trace_knowledge
     def query_knowledge(self, **kwargs) -> List[Document]:
@@ -241,7 +241,7 @@ class Knowledge(ComponentBase):
         for query_task in query_tasks:
             futures.append((
                 self.query_executor.submit(
-                    StoreManager().get_instance_obj(query_task[1]).query,
+                    StoreManager().get_instance_obj(query_task[1], strict=True).query,
                     query_task[0]),
                 query_task[1],
             ))
--- a/agentuniverse/base/component/component_manager_base.py
+++ b/agentuniverse/base/component/component_manager_base.py
@@ -47,17 +47,40 @@ class ComponentManagerBase(Generic[ComponentTypeVar]):
         self._instance_obj_map.pop(component_instance_name)
 
     def get_instance_obj(self, component_instance_name: str,
-                         appname: str = None, new_instance: bool = True) -> ComponentTypeVar:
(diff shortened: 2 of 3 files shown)


hxbszyy (changes requested):
The strict lookup is useful for #207, but this branch cannot be merged as-is. It conflicts with the channel-aware recall logic now on master from #609; the resolution must preserve the `(future, store_code)` pairs, channel metadata, and the opt-in deduplication contract while adding `strict=True`. Please rebase and add coverage that exercises strict lookup with channel-aware fusion enabled. Also remove the claim that this fixes or closes #203: that failure happened before a `Knowledge` instance existed because the caller requested `account_rule_knowledge` while the YAML registered `account_case_knowledge`; this patch only adds strict lookup inside an already-resolved `Knowledge` object.

123123213weqw:
@hxbszyy Thanks — both points are addressed.

1. Rebased onto current master (which now carries #609 channel-aware recall). The strict=True lookup is preserved alongside the (future, store_code) pairs, the recall_channel metadata, and the opt-in deduplication contract — the query_knowledge loop keeps its tuple/fusion shape and only the get_instance_obj call gains strict=True. Added TestStrictLookupWithChannelFusion, which runs the real query_knowledge with an RRF fusion processor configured: it asserts the stores are resolved via strict=True and that the shared document still fuses to 2/61 across both channels (strict and channel-aware fusion coexist). test_knowledge_rrf_integration still passes unchanged.

2. Dropped the #203 framing from the title and body. #203 is a name mismatch (account_rule_knowledge vs the registered account_case_knowledge) that fails before a Knowledge instance exists; this patch only adds strict lookup inside an already-resolved Knowledge, so it targets #207 and does not claim to fix #203.

`python -m unittest test_knowledge_registration_errors` -> 6 passed.

hxbszyy (approved):
The revision addresses both blockers cleanly. It is now based on the channel-aware retrieval path, preserves the future/store pairing and fusion semantics, and adds an integration case that exercises strict lookup with RRF enabled. The #203 claim is also gone and the scope is correctly limited to #207. I tested the new registration-error suite together with the existing RRF integration suite on current master: 11 passed. Approving.

Issue #258, opened 2025-01-16
Enhancement for knowledge pre-processing components/知识加工组件贡献


LandJerry:
======================== English version ==============================

The agentUniverse currently provides default knowledge pre-processing components, and the project requires other processing methods. We look forward to the following directions for PR:

Knowledge Fragmentation: Offer the most suitable knowledge fragmentation component based on specific task scenarios. For instance, in knowledge processing related to contract clauses, it is often necessary to fragment based on individual clauses; in academic paper processing, fragmentation might be based on arguments and evidence.

Knowledge Extraction: Provide the most appropriate knowledge extraction component based on specific task scenarios. For example, in supply chain knowledge scenarios, it is often crucial to extract entities and their relationships; in financial report scenarios, the extraction of corresponding financial indicators is typically needed.

Knowledge Deduplication: Offer the most suitable knowledge deduplication component based on specific task scenarios, including knowledge aggregation, summarization, and synthesis. For instance, in financial event scenarios, it is necessary to aggregate associated news articles related to the same event into a single event report.

======================== 中文版 ==============================

agentUniverse目前提供默认的知识加工组件，项目需要其他的加工手段，期待您建设的方向如下：

* 知识拆条：需要结合具体任务场景提供最合适的知识拆条组件。例如在合同条款类知识加工中，往往按照单条条款进行拆条；在论文类知识加工中，往往按照论点、论据进行拆条；
* 知识提取：需要结合具体任务场景提供最合适的知识拆条组件。例如，在供应链场景知识中往往会提取知识中的实体与其关联关系；在金融财报场景知识中往往会提取对应的金融指标；
* 知识去重：需要结合具体任务场景提供最合适的知识去重组件，包括对于知识的聚合、摘要、总结。例如，在金融事件场景中，需要将同一个事件的关联新闻聚合到同一个事件中去。

SeasonPilot:
Please take a look at whether this plan is feasible. If there are no issues, I will start implementing it.


## Technical Requirements

### 1. Knowledge Fragmentation (知识拆条)

**Objective**: Provide context-aware knowledge fragmentation components tailored to specific task scenarios.

**Use Cases**:
- **Contract Processing**: Fragment by individual clauses
  - Each legal clause treated as independent knowledge unit
  - Maintain clause hierarchy and relationships
  - Preserve legal context and dependencies

- **Academic Paper Processing**: Fragment by arguments and evidence
  - Separate thesis statements from supporting evidence
[...]

PR #625, opened 2026-07-13
feat(knowledge): MarkdownHeaderTextSplitter pre-processing component (#258)

123123213weqw:
## What

A new knowledge **pre-processing** `DocProcessor` — `MarkdownHeaderTextSplitter` — for the *knowledge pre-processing components* direction of #258. (Sibling of the post-processing `ReciprocalRankFusionProcessor` in #609: that one fuses recall *after* retrieval, this one shapes docs *before* embedding.)

It splits each input `Document` into one section per markdown ATX header (`#` … `######`), and records each section header hierarchy as metadata so a retrieved chunk can be traced back to where it came from:

```
# Installation
## macOS
brew install ...
```
→ two chunks, the second carrying `metadata["header_path"] = "Installation > macOS"`.

## Why

Long markdown sources (`.md` files, crawled web pages, READMEs) embed poorly as a single blob. Header-based chunking is a standard RAG preprocessing step and was missing from the framework. Unlike the existing splitters (`character` / `recursive_character` / `token` / `hierarchical_regex`), this one is **pure-Python with no third-party dependency** — it does not wrap a langchain splitter — so it is fully unit-testable without a network connection or optional install, and is safe to use even where `langchain` is not installed.

It only separates on headers (its single responsibility); a section that is still too large should be chained with a character/token splitter afterwards — splitter processors compose, the output list is a valid input to the next one.

## How it works

- ATX headers matched by `^(#{1,6})[ \t]+(.+?)[ \t]*#*$` (recognizes closed-ATX `## Title ##`; `#NoSpace` and 7+ hashes are content, not headers).
- An open-header stack is maintained; a header at level `L` closes any deeper open headers, mirroring markdown outline semantics (sibling headers replace, parent headers reset children).
- Preamble (text before the first header) is emitted as its own section by default, or droppable via `keep_preamble: false`.
- Each section preserves the source document existing metadata and adds the header path under a configurable key.

Configurable via the shipped yaml: `max_header_level` (1–6), `header_path_key`, `keep_preamble`, `path_separator`.

The yaml lives next to the module and is auto-registered through the default scanned package `agentuniverse.agent.action.knowledge.doc_processor` (same convention as the other built-in processors).

## Files

- `agentuniverse/agent/action/knowledge/doc_processor/markdown_header_text_splitter.py`
- `agentuniverse/agent/action/knowledge/doc_processor/markdown_header_text_splitter.yaml`
- `tests/.../doc_processor/test_markdown_header_text_splitter.py`

## Tests

13 tests, all deterministic, no network / no optional deps:
- 11 splitting-logic cases: header hierarchy, level reset, sibling replacement, preamble keep/drop, `max_header_level` limiting, closed-ATX, non-header hash lines, empty / no-header input, existing-metadata preservation, `header_path_key=None`.
- 2 config/registration cases through the real framework loader (`Configer → ComponentConfiger`): the shipped yaml resolves to `DOC_PROCESSOR` with the correct module/class — the component-schema contract.

`pytest test_markdown_header_text_splitter.py` → **13 passed**.

Relates to #258.

### DIFF
--- a/agentuniverse/agent/action/knowledge/doc_processor/markdown_header_text_splitter.py
+++ b/agentuniverse/agent/action/knowledge/doc_processor/markdown_header_text_splitter.py
@@ -22,6 +22,10 @@ fully unit-testable without a network connection or optional install.
 It only separates on headers; a section that is still larger than desired
 should be chained with a character / token splitter afterwards (splitter
 processors compose — the output list is a valid input to the next one).
+
+Fenced code blocks (backtick or tilde fences) are respected: a '#'-prefixed
+line inside one is treated as code rather than a header, so code samples in a
+README are not mistaken for section titles.
 """
 
 import re
@@ -38,6 +42,16 @@ from agentuniverse.base.config.component_configer.component_configer import \
 # trailing run of '#' (closed ATX syntax) and surrounding whitespace are
 # stripped. A line such as "#no-space" or "####### seven" is not a header.
 _HEADER_RE = re.compile(r"^(#{1,6})[ \t]+(.+?)[ \t]*#*[ \t]*$")
+
+# Fenced code blocks (CommonMark). An opening fence is up to three leading
+# spaces followed by three or more backticks or tildes; a backtick fence may
+# not carry a backtick in its info string. A closing fence is up to three
+# leading spaces, the same fence character repeated at least as many times as
+# the opening fence, then only trailing whitespace. While inside a fence, ATX
+# header detection is suppressed so a '# Example' line inside a ```python
+# block is treated as code instead of a header.
+_FENCE_OPEN_RE = re.compile(r"^( {0,3})(`{3,}|~{3,})(.*)$")
+_FENCE_CLOSE_RE = re.compile(r"^( {0,3})(`{3,}|~{3,})[ \t]*$")
 
 
 class MarkdownHeaderTextSplitter(DocProcessor):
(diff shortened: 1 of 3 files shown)


hxbszyy (changes requested):
The component is useful and fits the preprocessing direction, but the parser currently treats ATX-looking lines inside fenced code blocks as real headings. A README containing a Python or shell example with a line such as `# Example` will be split and assigned the wrong header hierarchy. Please track opening and closing backtick and tilde fences, suppress heading recognition while inside them, and add regression tests for both fence styles before this lands.

123123213weqw:
@hxbszyy Thanks — addressed in the latest commit. The splitter now tracks fenced code blocks (backtick and tilde) per CommonMark: it follows opening and closing fences, requires the closing fence to use the same character with a run at least as long as the opening, and does not treat a backtick info string containing a backtick as an opening fence. Header recognition is suppressed while inside a block, so a `# Example` line in a ```python snippet stays as code. Added regression tests for both fence styles, fence-char mismatch (tilde will not close a backtick fence), the run-length rule, and an unclosed fence. `python -m unittest ...test_markdown_header_text_splitter` -> 19 passed.

hxbszyy (approved):
The fenced-code handling now covers the important CommonMark cases: both fence characters, matching close character, minimum closing length, invalid backtick info strings, and unclosed fences. That removes the correctness issue for code-heavy Markdown while keeping the component focused on header-aware preprocessing. I ran the full splitter suite on current master: 19 passed. Approving.

PR #626, opened 2026-07-13
feat(knowledge): LengthFilter pre-processing component (#258)

123123213weqw:
## What

A new knowledge **pre-processing** `DocProcessor` — `LengthFilter` — for #258. It drops documents whose text length falls outside a configurable `[min_length, max_length]` range. This is the standard RAG hygiene step to chain right after a splitter: splitters (`character` / `token` / the `MarkdownHeaderTextSplitter` in #625) frequently emit near-empty fragments — a header whose body wrapped to the next line, a stray blank section — and those fragments embed to noise and pollute retrieval.

## Why it does not overlap with existing components

The framework already has `threshold_filter`, but that filters by **relevance score *after* recall**. `LengthFilter` filters by **content length *before* insert**. Different axis, different stage — they compose cleanly:

```
reader → markdown_header_text_splitter → length_filter → store
```

## How

Pure-Python, dependency-free. Length is measured in **characters** by default, or in **whitespace-separated tokens** when `length_unit: token` (an approximate, dependency-free token count — not a model tokenizer). `trim: true` (default) strips leading/trailing whitespace before measuring so a whitespace-only document counts as length 0.

It is a **filter, not a transformer**: kept documents are passed through unchanged, so their ids and metadata are preserved.

Configurable via the shipped yaml: `min_length` (default 1 — removes empty / whitespace-only out of the box), `max_length` (default null = no upper bound), `length_unit`, `trim`.

## Files

- `agentuniverse/agent/action/knowledge/doc_processor/length_filter.py`
- `agentuniverse/agent/action/knowledge/doc_processor/length_filter.yaml`
- `tests/.../doc_processor/test_length_filter.py`

## Tests

11 deterministic tests, no network / no optional deps:
- 9 logic cases: default drops empty/whitespace-only, min threshold, max threshold, combined bounds, token unit, `trim=false` counts raw whitespace, ids/metadata preserved (same object), empty input, `min=max=None` keeps everything.
- 2 config/registration cases through the real loader (`Configer → ComponentConfiger`): the shipped yaml resolves to `DOC_PROCESSOR` with the correct module/class.

`pytest test_length_filter.py` → **11 passed**.

Relates to #258.

### DIFF
--- a/agentuniverse/agent/action/knowledge/doc_processor/length_filter.py
+++ b/agentuniverse/agent/action/knowledge/doc_processor/length_filter.py
@@ -1,0 +1,107 @@
+# !/usr/bin/env python3
+# -*- coding:utf-8 -*-
+
+# @Time    : 2026/07/13
+# @FileName: length_filter.py
+
+"""
+Length-based document filter.
+
+Drops input :class:`Document` objects whose text length falls outside a
+configured ``[min_length, max_length]`` range. This is a *pre-processing*
+hygiene step for issue #258: a splitter (character, token, or the
+:class:`MarkdownHeaderTextSplitter`) frequently emits near-empty fragments —
+a header line whose body was on the next line, a stray blank section — and
+those fragments embed to noise and pollute retrieval. Chaining a length filter
+right after the splitter removes them before they reach the store.
+
+The existing ``threshold_filter`` filters by *relevance score* after recall;
+this filter operates by *content length* before insert, so the two are
+complementary and never overlap.
+
+Pure-Python and dependency-free. Length is measured in characters by default;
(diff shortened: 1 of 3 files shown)


hxbszyy:
Thanks for putting this together, but I do not think this should become a framework component. Issue #258 asks for scenario-aware fragmentation, extraction, and deduplication; a generic min/max length predicate does not address those capabilities, and callers can already apply this check in a pipeline without adding another registered component. The `token` mode is also whitespace word counting rather than token counting, which gives the option a misleading contract. The maintenance surface is larger than the value here, so I am closing this PR rather than asking for more polish.

PR #627, opened 2026-07-14
fix(tool): parse run command blocking flag

Loyal-Young:
**When submitting a PR, please confirm the following points and put [x] in the boxes one by one.** | **在提出pr时，请确认了以下几点，并逐一使用[x]符号确认为选。**

**Checklist | 检查项**
- [x] I have read and understood the [contributor guidelines](https://github.com/antgroup/agentUniverse/blob/master/CONTRIBUTING.md). | 我已阅读并理解[贡献者指南](https://github.com/antgroup/agentUniverse/blob/master/CONTRIBUTING_zh.md)。
- [x] I have submitted the test files and can provide screenshots of the test results (required for feature or bug fixes) | 我已经提交了测试文件并可提供测试结果截图(功能修改、BUG修复类PR必须提供，其他按需)

- Fix `RunCommandTool` parsing for the `blocking` option when it is passed through `ToolInput`.
- Previously, a string value such as `"false"` was treated as truthy by Python, so the command still ran in blocking mode.
- This PR normalizes common boolean strings such as `true/false`, `1/0`, `yes/no`, and `on/off` before command execution.

- `tests/test_agentuniverse/unit/agent/action/tool/test_run_command.py`
- `python -m py_compile agentuniverse/agent/action/tool/common_tool/run_command_tool.py tests/test_agentuniverse/unit/agent/action/tool/test_run_command.py`
- `git diff --check`

- None.

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/run_command_tool.py
+++ b/agentuniverse/agent/action/tool/common_tool/run_command_tool.py
@@ -16,6 +16,7 @@ from typing import Dict, Optional
 from dataclasses import dataclass
 
 from agentuniverse.agent.action.tool.tool import Tool, ToolInput
+from agentuniverse.agent.action.tool.common_tool.tool_input_utils import parse_strict_bool
 
 
 class CommandStatus(Enum):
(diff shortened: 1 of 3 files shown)


hxbszyy (changes requested):
The compatibility goal is useful, but this parser still accepts every unknown value. For example, `blocking: "flase"` falls through to `bool(value)` and becomes `True`. That turns malformed model or config output into a valid execution mode instead of reporting an input error. Please make coercion closed over the explicitly supported spellings and reject everything else, with a typo regression test. Since #628 needs the same rule, this should be one shared parser so the two tools cannot drift.

Loyal-Young:
Updated as suggested. Both tools now use the same strict boolean parser. Unknown strings such as "flase" and non-0/1 numeric values are rejected instead of falling back to Python truthiness. I also added regression tests for typo and invalid numeric inputs.

hxbszyy (approved):
This now fails closed on malformed values and shares one parser with the file-writing path instead of duplicating coercion rules. The typo and non-binary numeric cases prove that an invalid model/tool payload cannot silently select an execution mode. I ran the command and status suites on current master: 10 passed. Approving.

PR #632, opened 2026-07-15
[Refactor][Reader] Introduce typed semantic reader exceptions and replace generic runtime errors

zhaohai666:
## Summary
This PR introduces a dedicated hierarchy of semantic exception classes for all data readers, replaces scattered generic built-in exceptions with domain-specific typed errors, and migrates raw print statements to standard logging calls.

Close #【fill related issue number】

### 1. New File: `reader_errors.py`
Define 5 layered semantic exception classes:
- `ReaderError`: Base abstract exception, contains mandatory `reader_name` attribute
- `ReaderLoadError`: Raised for missing local file / remote HTTP fetch failures, carries `source` and optional `status_code`
- `ReaderDependencyError`: Raised when optional third-party library dependency is missing, carries `dependency` and actionable `install_hint`
- `ReaderParseError`: Raised when content parsing logic fails, carries `source` reference
- `ReaderConfigError`: Raised for invalid / incomplete reader configuration arguments, carries `config_key`

### 2. Refactor 11 reader implementation files
- Replace bare `ValueError`, `ImportError`, `FileNotFoundError`, `ConnectionError`, `RuntimeError` with corresponding typed reader exceptions
- Remove all raw `print()` logging, migrate to standard `logging` module calls
- Attach required context parameters (`reader_name`, source path, config key, dependency info etc.) when raising exceptions

### 3. New Unit Test: `test_reader_errors.py`
Total 30 unit test cases covering:
- Exception inheritance hierarchy validation
- Required attribute initialization & accessibility
- Human-readable exception message formatting
- Module public export list verification
- Attribute boundary & optional field behavior

### 4. New Integration Test: `test_reader_exception_paths.py`
Add end-to-end exception path integration tests across multiple readers to verify correct error types are thrown under failure scenarios:
- `JsonReader`: Missing file → `ReaderLoadError`; malformed JSON content → `ReaderParseError`
- `CSVReader`: Invalid construction arguments → `ReaderConfigError`
- `WebPageReader`: Empty / invalid URL supplied → `ReaderConfigError`
- `RenderedWebPageReader`: Missing playwright installation → `ReaderDependencyError`
- `PdfReader`: Missing pypdf dependency → `ReaderDependencyError`

## Purpose & Benefits
1. Enables upstream caller to safely differentiate failure categories without parsing exception message strings
2. Standardizes error context metadata across all readers for consistent logging & error handling
3. Provides actionable installation hints for missing optional dependencies
4. Unifies logging practice and eliminates ad-hoc print debugging
5. Establishes stable typed error contract for downstream consumers

## Test Status
- All 30 unit tests in `test_reader_errors.py` passed
- All reader exception path integration tests in `test_reader_exception_paths.py` passed

### DIFF
--- a/agentuniverse/agent/action/knowledge/reader/__init__.py
+++ b/agentuniverse/agent/action/knowledge/reader/__init__.py
@@ -5,3 +5,19 @@
 # @Author  : wangchongshi
 # @Email   : wangchongshi.wcs@antgroup.com
 # @FileName: __init__.py
+
+from agentuniverse.agent.action.knowledge.reader.reader_errors import (
+    ReaderError,
+    ReaderLoadError,
+    ReaderDependencyError,
+    ReaderParseError,
+    ReaderConfigError,
+)
+
+__all__ = [
+    "ReaderError",
+    "ReaderLoadError",
+    "ReaderDependencyError",
+    "ReaderParseError",
+    "ReaderConfigError",
+]
--- a/agentuniverse/agent/action/knowledge/reader/file/csv_reader.py
+++ b/agentuniverse/agent/action/knowledge/reader/file/csv_reader.py
@@ -7,12 +7,20 @@
 
 import csv
 import io
+import logging
 from pathlib import Path
 from typing import List, Union, Optional, Dict, TextIO
 
 from agentuniverse.agent.action.knowledge.reader.reader import Reader
+from agentuniverse.agent.action.knowledge.reader.reader_errors import (
+    ReaderLoadError,
+    ReaderConfigError,
+    ReaderParseError,
+)
 from agentuniverse.agent.action.knowledge.store.document import Document
 from agentuniverse.agent.action.knowledge.reader.utils import detect_file_encoding
+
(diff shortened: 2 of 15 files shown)


hxbszyy (changes requested):
A typed reader hierarchy is useful, but this migration breaks the existing public exception contract. Callers that currently catch `FileNotFoundError`, `ImportError`, `ValueError`, or `json.JSONDecodeError` will stop catching the same failures because every new type inherits only from `ReaderError`. The compatibility test only checks `Exception`, which is not meaningful compatibility. Please preserve the relevant built-in catch paths, or introduce the hierarchy without changing existing raised types and migrate in a staged release. The branch also fails its own focused suite: 45 passed, 1 skipped, and the Docx dependency test failed because the reader uses `docx2txt` while the test expects `python-docx`. Existing JSON tests now fail on the changed exception types, and the new XLSX existence check breaks mocked success paths. These regressions need to be resolved before a subsystem-wide exception rewrite can land.

PR #633, opened 2026-07-15
fix(config): preserve explicit component configuration values

messere1:
**Checklist | 检查项**

- [x] I checked open issues and pull requests for duplicate work; no matching work was found.
- [x] I accept maintainer suggestions to change or close this PR.
- [x] I have submitted regression tests for these bug fixes.
- [ ] Documentation changed (not needed; internal configuration behavior only).
- [ ] Examples added (not needed for these regression fixes).

## Summary

- Use the inherited public `configer` property in eight component configurer `load()` methods, avoiding Python name-mangling `AttributeError` failures.
- Preserve explicit falsy LLM settings: `temperature=0.0`, `max_retries=0`, `streaming=False`, and empty `ext_info`.
- Load Planner `input_key`, `output_key`, and `memory_key` instead of silently leaving them unset.

## Motivation

Subclass loaders accessed `self.__configer` even though that private attribute belongs to `ComponentConfiger`, so direct `load()` calls fail due to Python name mangling. Truthiness checks also discarded valid falsy LLM settings, while PlannerConfiger declared data-key fields without loading them.

## Tests

- `tests/test_agentuniverse/unit/llm/test_llm_configuration.py`
- `tests/test_agentuniverse/unit/base/config/test_component_configer_load.py`
- `tests/test_agentuniverse/unit/base/config/test_planner_configer.py`

```text
python -m pytest tests/test_agentuniverse/unit/llm/test_llm_configuration.py tests/test_agentuniverse/unit/base/config/test_component_configer_load.py tests/test_agentuniverse/unit/base/config/test_planner_configer.py -q
# 12 passed

ruff check --no-fix <the three test files>
# All checks passed

black --check <the three test files>
# 3 files would be left unchanged
```

## Compatibility

No new dependencies or public API changes. `None` continues to mean “do not override”; explicit falsy values are now respected.

### DIFF
--- a/agentuniverse/base/config/component_configer/configers/knowledge_configer.py
+++ b/agentuniverse/base/config/component_configer/configers/knowledge_configer.py
@@ -45,7 +45,7 @@ class KnowledgeConfiger(ComponentConfiger):
         Returns:
             KnowledgeConfiger: the KnowledgeConfiger object
         """
-        return self.load_by_configer(self.__configer)
+        return self.load_by_configer(self.configer)
 
     def load_by_configer(self, configer: Configer) -> 'KnowledgeConfiger':
         """Load the configuration by the Configer object.
--- a/agentuniverse/base/config/component_configer/configers/llm_configer.py
+++ b/agentuniverse/base/config/component_configer/configers/llm_configer.py
@@ -80,7 +80,7 @@ class LLMConfiger(ComponentConfiger):
         Returns:
             LLMConfiger: the LLMConfiger object
         """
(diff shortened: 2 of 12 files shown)


hxbszyy (approved):
This fixes three real configuration bugs rather than papering over symptoms: direct subclass loading now uses the inherited configer, explicit falsy LLM values survive both initialization and agent overrides, and planner data keys are actually loaded. I ran the focused and adjacent config suite on current master in the agentUniverse environment: 21 passed. Approving.

PR #634, opened 2026-07-16
[Refactor][Reader] Directory restructuring, cloud reader consolidation and ReaderManager capability enhancement

zhaohai666:
## Summary
This PR reorganizes cloud reader module structure, consolidates cloud document readers, rewrites FeishuReader, migrates YuqueReader, and enhances ReaderManager with URL pattern routing capability. Meanwhile, 40 stub-based unit tests are standardized and migrated.

Close #【Related Issue Number】

### 1. Directory Restructuring
- Move directory `cloud_file_reader/` → `cloud/` as the new root namespace for all cloud document readers

### 2. Unified Cloud Entry & URL Domain Routing
- Introduce `CloudDocReader` as unified top-level entry for all cloud document resources
- Implement domain-based URL routing dispatching inside CloudDocReader to match corresponding concrete reader implementation

### 3. Full Rewrite of FeishuReader
- Adopt lazy-loading WebDriver initialization to reduce cold startup overhead
- Integrate newly defined semantic reader exceptions for all failure branches
- Add `PublicFeishuReader` as compatible type alias

### 4. YuqueReader Migration & Refactor
- Migrate YuqueReader into the new `cloud/` module directory
- Refactor all error paths to raise standardized semantic reader exceptions

### 5. ReaderManager Enhancement
- Add static `URL_PATTERN_MAP` registry storing domain → reader mapping rules
- Implement new public method `get_url_default_reader()` to automatically resolve matched reader class by input URL

### 6. Test Refactoring
- Total 40 original stub-based unit tests converted to formal standardized test cases
- Cover URL matching logic, lazy initialization flow, exception propagation, reader alias compatibility and manager dispatching behavior

## Benefits
1. Cleaner module hierarchy, unified management of all cloud document readers under `cloud/` namespace
2. Single entry `CloudDocReader` simplifies upstream usage; automatic URL dispatching reduces caller manual reader selection
3. Lazy WebDriver initialization optimizes resource consumption for Feishu document access
4. All cloud readers fully aligned with global typed exception specification, consistent error handling contract
5. ReaderManager gains built-in URL auto-detection capability for end-user convenience
6. Test suite upgraded from placeholder stubs to validated functional test coverage

## Test Result
All 40 migrated unit tests pass.
Exception error paths, URL pattern matching and lazy initialization behaviors are fully verified.

PR #635, opened 2026-07-16
Feature/reader dependency groups

zhaohai666:
## Summary
This PR reorganizes reader optional dependencies and extras definitions in `pyproject.toml`, fixes module export in cloud reader entry, and replaces raw print statements with standard logger calls in website reader.

Close #【Related Issue Number】

### 1. pyproject.toml Dependency Group Refactoring
- Declare total **18 optional reader individual dependencies**
- Define **13 categorized extras groups** for convenient installation:
  `pdf`, `docx`, `pptx`, `web-extract`, `web-render`, `ocr-paddle`, `ocr-tesseract`, `ocr-easy`, `ocr`, `pdf-ocr`, `selenium-reader`, `cloud`, `reader-all`
- Each extras group aggregates related optional packages to simplify end-user installation commands

### 2. cloud/__init__.py Module Export Fix
- Add explicit export declaration for `PublicFeishuReader` to expose the alias in cloud module public API

### 3. website_bs4_reader.py Logging Standardization
- Replace all bare `print()` warning outputs with standardized `logger.warning()` calls
- Unify logging practice consistent with other reader components

## Benefits
1. Clearer optional dependency management, users can install only required reader subsets via extras tags
2. Complete public API exposure for cloud reader aliases without import obstacles
3. Remove ad-hoc print debugging output, align all readers to unified logging specification

## Test Result
- Dependency group definition syntax verified
- Module import test for `PublicFeishuReader` passed
- Log output paths validated after replacing print statements


hxbszyy (changes requested):
The dependency grouping is a useful direction, but this branch is not installable. `docx2txt = ">=0.36"` matches no published release; PyPI currently tops out at 0.9, and running `poetry lock --no-interaction` fails with “docx2txt (>=0.36) ... does not match any versions.” Please correct and verify every constraint by resolving the lock and installing each extra in a clean environment, not just by checking that TOML keys reference each other. This PR also repeats all of #632 and #634 on top of master, so it inherits their exception-compatibility and removed-import-path blockers. Retarget or rebuild it as a clean follow-up after those bases are accepted.

zhaohai666:
> 依赖项分组是一个有用的方向，但此分支无法安装。`docx2txt = ">=0.36"`它与任何已发布的版本都不匹配；PyPI 目前最高版本为 0.9，运行`poetry lock --no-interaction`失败并显示“docx2txt (>=0.36) ... 不匹配任何版本”。请通过解决锁定并在干净的环境中安装每个额外的依赖项来修正并验证每个约束，而不仅仅是检查 TOML 键是否相互引用。此 PR 还重复了#632和#634的所有内容，并基于 master 分支，因此它继承了它们的异常兼容性和移除导入路径的阻塞问题。在这些基础分支被接受后，请将其重新定位或重新构建为一个干净的后续分支。

It has been fixed, take another look

hxbszyy:
Thanks for fixing the constraint. I checked the new head: `docx2txt >=0.8` now resolves, and `poetry lock`, `poetry check`, and a dry-run install of `reader-all` all complete successfully. That clears the dependency-version blocker.

The branch still is not ready to merge, though. It still carries the complete #632 and #634 commit stacks (33 files and roughly 2.7k added lines), including the exception-compatibility and reader-path blockers from those reviews. The new commit did not split or rebase that work away. Please rebuild this as a clean PR from current master containing only the dependency/extras changes (plus any directly necessary tests). Once the branch is actually isolated, I will verify the individual extras against that clean diff.

PR #636, opened 2026-07-16
fix(core): preserve async compatibility and task isolation

messere1:
## Summary

- preserve deprecated `ToolInput` execution in `async_run` without blocking the event loop
- make `async_langchain_run` match the legacy synchronous ReAct-input behavior
- return context tokens from `set_all_contexts` so executing-agent workers restore their previous context
- isolate `InputObject` from constructor and `to_dict()` mapping mutations

## Why

These fixes close three runtime state/compatibility gaps that can surface under asynchronous or parallel agent execution: legacy tools fail only on async paths, worker context leaks after subtask execution, and copied input objects can still share mutable mapping state.

## Tests

- `python -m pytest tests/test_agentuniverse/unit/agent/action/tool/test_tool_async_compatibility.py tests/test_agentuniverse/unit/agent/test_input_object.py -q` (4 passed)
- `python -m pytest tests/test_agentuniverse/unit/base/context/test_framework_context.py -k set_all_contexts_returns_tokens_for_restoration -q` (1 passed)
- `python -m py_compile agentuniverse/agent/action/tool/tool.py agentuniverse/agent/input_object.py agentuniverse/agent/template/executing_agent_template.py agentuniverse/base/context/framework_context_manager.py`
- `git diff upstream/master...HEAD --check`

No new dependencies are introduced.

### DIFF
--- a/agentuniverse/agent/action/tool/tool.py
+++ b/agentuniverse/agent/action/tool/tool.py
@@ -82,6 +82,8 @@ class Tool(ComponentBase):
     async def async_run(self, **kwargs):
         """The callable method that runs the tool."""
         self.input_check(kwargs)
+        if self.check_execute_signature_deprecated():
+            return await asyncio.to_thread(self.execute, ToolInput(kwargs))
         return await self.async_execute(**kwargs)
 
     def input_check(self, kwargs: dict) -> None:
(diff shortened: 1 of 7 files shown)


hxbszyy (approved):
This fixes three real runtime issues without widening the public surface: legacy ToolInput-based tools now keep working through async entry points, worker threads restore their inherited framework context, and InputObject no longer exposes its internal mapping by reference. I rebased the change locally onto the current master and reran the focused async/context/input tests, including the previously skipped async isolation test; all 6 passed.

Looks good to merge.

PR #638, opened 2026-07-16
feat(knowledge): MMRProcessor for diversity-aware re-ranking (#248)

123123213weqw:
## What

Adds `MMRProcessor`, a knowledge *post-processor* (#248) that re-ranks recalled documents with **Maximal Marginal Relevance (MMR)** so a result set is both on-topic and non-repetitive.

At each step it selects the document maximising:

```
score(d) = lambda * sim(d, query) - (1 - lambda) * max_{selected} sim(d, selected)
```

- `lambda_coef = 1.0` → pure relevance ranking
- `lambda_coef = 0.0` → maximise diversity
- `lambda_coef = 0.5` (default) → balance

## Why this component

Fills a gap in the current `doc_processor` set (verified: no MMR today), and is intentionally distinct from the siblings it might be compared to, to avoid redundancy:

- vs **`semantic_deduplicator`** — that *removes* near-duplicates above a hard threshold; MMR does soft, diversity-aware *selection / re-ranking* with a relevance/diversity trade-off.
- vs **`ReciprocalRankFusionProcessor`** — that *combines* ranked lists from several stores (cross-channel score fusion); MMR re-ranks a single recalled set for intra-list diversity.

Both the similarity matrix and the greedy selection make it non-trivial (not a few-line inline), complementing the existing rerank / dedup / fusion components.

## Design

- **Embeddings**: query/doc vectors are read from `Query.embeddings` / `Document.embedding` (the store populates them during retrieval). When `embedding_name` is set, missing embeddings are computed on demand via `EmbeddingManager` (same pattern as `semantic_deduplicator`).
- **Graceful fallback**: if embeddings cannot be obtained, the documents are returned in input order instead of crashing retrieval.
- **No score clobbering**: `score_key` defaults to empty, so MMR does not overwrite `relevance_score` stamped by earlier processors such as RRF; opt in by setting `score_key`.
- Pure-Python cosine (dependency-free, fully unit-testable).

## Config

```yaml
name: 'mmr_processor'
metadata:
  type: 'DOC_PROCESSOR'
  module: 'agentuniverse.agent.action.knowledge.doc_processor.mmr_processor'
  class: 'MMRProcessor'
lambda_coef: 0.5
top_n: null
embedding_name: ''
score_key: ''
```

## Tests

`tests/.../test_mmr_processor.py` — 19, all pass:
- **Selection**: `lambda=1` → relevance order; `lambda=0` → diversity order (an orthogonal document is promoted over a redundant one); `top_n` truncation; most-relevant-first regardless of lambda; `score_key` stamps cosine relevance.
- **Embeddings**: uses `Query.embeddings` directly (no model call); falls back to input order when embeddings are missing and no `embedding_name`; `embedding_name` computes missing embeddings; an embedding-lookup failure degrades safely.
- **Config**: `lambda_coef` validated to `[0, 1]`; attributes loaded from the configer.
- **Integration**: re-ranks through the real `Knowledge.query_knowledge` → `post_processors` path.

`python -m unittest test_mmr_processor` → 19 passed; `test_knowledge_rrf_integration` still passes (no regression).

Addresses the re-ranking / diversity direction of #248.

### DIFF
--- a/agentuniverse/agent/action/knowledge/doc_processor/mmr_processor.py
+++ b/agentuniverse/agent/action/knowledge/doc_processor/mmr_processor.py
@@ -102,6 +102,20 @@ class MMRProcessor(DocProcessor):
                 "returning documents in input order.")
             return origin_docs[:self.top_n] if self.top_n else list(origin_docs)
 
+        # Guarantee a homogeneous embedding space before ranking. Documents
+        # recalled from multiple stores can carry vectors from different models
+        # or dimensions; comparing them with cosine would silently truncate to
+        # the shortest vector and produce a plausible but meaningless ranking.
+        # Require one shared dimension, else degrade to input order.
+        dims = {len(e) for e in doc_embeddings} | {len(query_embedding)}
+        if len(dims) > 1:
+            logger.warning(
+                f"MMR: embeddings have inconsistent dimensions {sorted(dims)}; "
+                "cannot compute meaningful similarities. Returning documents in "
+                "input order. Set embedding_name to recompute all embeddings "
+                "with a single model.")
+            return origin_docs[:self.top_n] if self.top_n else list(origin_docs)
+
         order = self._select(doc_embeddings, query_embedding)
         if self.score_key:
             for i in order:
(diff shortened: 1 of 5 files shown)


hxbszyy (changes requested):
MMR is a useful and distinct processor, but the current embedding handling can silently rank on invalid similarities. `_cosine` uses `zip(a, b)` without requiring equal dimensions, and `_resolve_embeddings` reuses every existing `Document.embedding` even when `embedding_name` is configured. With multi-store retrieval, documents can carry vectors from different models or dimensions; the code will silently truncate mismatched vectors and produce a plausible-looking but meaningless ranking.

Please guarantee a homogeneous embedding space before MMR runs. A configured `embedding_name` should recompute the query and all document embeddings with that model, or the processor must otherwise validate dimension and provenance and fail/degrade explicitly. Please add mixed-dimension and mixed precomputed/computed embedding tests as well.

123123213weqw:
Thanks @hxbszyy — addressed in db61bacf.

**1. `_cosine` no longer silently truncates.** It now raises `ValueError` when the two vectors differ in length, instead of zipping to the shorter one.

**2. A configured `embedding_name` recomputes everything.** `_resolve_embeddings` now branches: when `embedding_name` is set, `_resolve_embeddings_via_model` recomputes **all** document embeddings *and* the query with that single model, **ignoring precomputed vectors** — so documents recalled from multiple stores (different models/dimensions) are forced into one homogeneous space. The old code reused every `Document.embedding` even with a model configured, which is exactly the multi-store mismatch you flagged.

**3. Explicit degradation on heterogeneous precomputed embeddings.** When no model is configured, `_process_docs` validates up front that every document embedding and the query embedding share one dimension; if not, it logs a warning and returns the documents in input order rather than ranking on truncated/garbage similarities. (`_cosine`'s raise is the invariant guard behind that.)

**Tests** (new `TestMMREmbeddingHomogeneity`):
- `test_cosine_rejects_mismatched_dimensions` — `_cosine([1,0],[1,0,0])` raises.
- `test_mixed_document_dimensions_degrade_to_input_order` — docs of dim 3 and dim 2 with no model degrade to input order.
- `test_query_document_dimension_mismatch_degrades_to_input_order` — query dim differs from docs → input order.
- `test_embedding_name_recomputes_all_ignoring_precomputed` — docs carry bogus precomputed vectors of *different* dims (`[9,9,9]`, `[1]`, `[0,0]`); with `embedding_name` set they are recomputed homogeneously, the precomputed ones are overwritten (`docs[0].embedding == _D1`), and MMR ranks by relevance `d1 > d2 > d3`.

All 23 tests in the file pass.

hxbszyy (changes requested):
Thanks for fixing the document-side mixing and the dimension truncation. One provenance hole remains before this is actually homogeneous.

When embedding_name is configured but Query.query_str is absent, _resolve_embeddings_via_model() falls back to Query.embeddings. Query.query_str is optional in the public model, and an equal-dimensional query vector from a different model will pass the new dimension check and still be compared against the recomputed document vectors. Equal dimensions do not imply the same embedding space.

With a configured model, please either recompute the query from text, verify explicit model provenance, or fail/degrade when only an unverifiable precomputed query vector is available. Add a regression case where the foreign query vector has the same dimension as the configured model so the dimension guard cannot hide the issue.

123123213weqw:
Thanks @hxbszyy — addressed in cabef37b.

The provenance hole is now closed. When `embedding_name` is configured, `_resolve_embeddings_via_model` embeds the query **only from `query_str`**; a query that carries only a precomputed vector is explicitly rejected (with a logged reason) and the processor degrades to input order — the precomputed vector no longer reaches `_cosine`, so equal-dimension cross-model contamination cannot happen.

```python
if query is not None and query.query_str:
    query_embedding = model.get_embeddings([query.query_str])[0]
elif query is not None and not query.query_str and query.embeddings:
    logger.warning(
        "MMR: embedding_name is configured but the query carries no "
        "query_str, only a precomputed vector. Its provenance cannot "
        "be verified against the configured model (equal dimensions "
        "do not imply the same embedding space); returning documents "
        "in input order. Pass query_str to embed the query with the "
        "configured model.")
```

**Regression test added** — `test_embedding_name_rejects_precomputed_query_without_query_str`: the foreign query vector has the **same dimension (2)** as the configured model would emit, so the dimension guard alone cannot mask the issue; the test asserts input-order fallback and that the model is never asked to embed a query string (only one `get_embeddings` call, for the documents).

Full suite: 23/24 pass; the one failure (`test_mmr_reranks_in_the_pipeline`) is a pre-existing local-environment `ModuleNotFoundError: No module named 'mcp'` on the import chain — it fails identically on `master` and is unrelated to this change.

hxbszyy (approved):
This closes the provenance gap I was worried about. A configured model now defines one embedding space for every document and for the query; an equal-dimensional but unverifiable precomputed query is no longer allowed to slip through. The mixed-dimension, mixed-origin, and pipeline cases all pass (24 focused tests), and after resolving the current documentation conflict the MMR, redaction, and context-budget processor suites pass together (74 tests). I also aligned the guidebook wording with the new recompute-all contract. This is a distinct, useful post-retrieval capability now. Approving.

PR #639, opened 2026-07-16
feat(knowledge): ContextBudgetCompressor for context-window fitting (#248)

123123213weqw:
## What

Adds `ContextBudgetCompressor`, a knowledge *post-processor* (#248) that fits recalled documents into a fixed cumulative size budget (typically the LLM context window). It walks the recalled list — already ranked by the store or an earlier reranker/fusion processor — keeping documents in order while their total size stays within `budget`; the boundary document that would overflow is optionally **truncated** so the budget is used fully without exceeding it.

## Why this component (distinct from ThresholdFilter)

`ThresholdFilter` applies **per-document** predicates (a score/length range) or a fixed top-k. This component manages a different axis — the **cumulative** size of the kept set — and can split the last document to fit. That is the standard "fit recall into the model's context window" job, which per-doc filtering cannot do.

## Honest counting (no "token = word count" trap)

Size is measured by an explicit `counter`, and the budget is always interpreted in that counter's unit, so the contract is never "tokens" while silently counting words:

| counter | size of a doc |
|---|---|
| `estimate` (default) | `max(1, len(text)//4)` — dependency-free token approximation |
| `tiktoken` | real BPE tokens via `tiktoken` (encoding configurable); falls back to `estimate` with a warning if tiktoken is absent |
| `char` | exact character count |
| `word` | exact whitespace word count |

## Config

```yaml
name: 'context_budget_compressor'
metadata:
  type: 'DOC_PROCESSOR'
  module: 'agentuniverse.agent.action.knowledge.doc_processor.context_budget_compressor'
  class: 'ContextBudgetCompressor'
budget: 4096
counter: 'estimate'
truncate: true
tiktoken_encoding: 'cl100k_base'
```

## Tests

`tests/.../test_context_budget_compressor.py` — 20, all pass:
- **Selection**: keeps whole docs within budget; stops at first overflow when `truncate=false`; truncates the boundary doc (and a single oversize doc) to fit when `truncate=true`; preserves input order; preserves the truncated doc's id and metadata and stamps `truncated: true`.
- **Counters**: char / word / estimate counts and truncation behave exactly; `tiktoken` counts real BPE tokens and truncates by token (skipped when tiktoken is not installed).
- **Config**: invalid `counter` raises; attributes load from the configer.
- **Integration**: compresses through the real `Knowledge.query_knowledge` → `post_processors` path.

`python -m unittest test_context_budget_compressor` → 20 passed.

Addresses the context-window-management direction of #248. Composes naturally after a reranker or `ReciprocalRankFusionProcessor` (which order the recall list this processor then trims).

### DIFF
--- a/agentuniverse/agent/action/knowledge/doc_processor/context_budget_compressor.py
+++ b/agentuniverse/agent/action/knowledge/doc_processor/context_budget_compressor.py
@@ -36,7 +36,7 @@ never "tokens" while silently counting words.
 """
 
 import logging
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 from agentuniverse.agent.action.knowledge.doc_processor.doc_processor import \
     DocProcessor
@@ -49,9 +49,11 @@ logger = logging.getLogger(__name__)
 
 _VALID_COUNTERS = {"estimate", "tiktoken", "char", "word"}
 
-# Module-level tiktoken cache so repeated processing does not reload it.
-_TIKTOKEN_ENCODER = None
-_TIKTOKEN_TRIED = False
+# Cache of tiktoken encoders keyed by encoding name. Keying by name (rather
+# than a single module-global slot) means multiple compressor instances can use
+# different encodings, and a failed lookup for one encoding never poisons
+# another. A value of None means "tiktoken not installed" for that name.
+_TIKTOKEN_ENCODERS: Dict[str, Optional[object]] = {}
 
 
 class ContextBudgetCompressor(DocProcessor):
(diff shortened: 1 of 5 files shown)


hxbszyy (changes requested):
The cumulative budget behavior is valuable, but the tiktoken cache currently makes configuration order affect correctness. `_TIKTOKEN_ENCODER` / `_TIKTOKEN_TRIED` are module-global and not keyed by `tiktoken_encoding`, so the first compressor instance fixes the encoder for every later instance. If the first encoding name is invalid, every later valid instance also silently falls back to the estimate.

Please cache encoders by encoding name (a small dict or `lru_cache` is enough), and distinguish an unavailable optional dependency from an invalid configured encoding. A bad encoding should raise a clear configuration error rather than quietly changing the counting unit. Add tests with two different encodings and an invalid encoding before a valid one.

123123213weqw:
Thanks @hxbszyy — addressed in 77309780.

**1. Cache keyed by encoding name.** Replaced the single module-global slot (`_TIKTOKEN_ENCODER`/`_TIKTOKEN_TRIED`) with `_TIKTOKEN_ENCODERS: Dict[str, Optional[object]]` keyed by encoding name. Each compressor instance now resolves its *own* encoder, and a lookup for one encoding never poisons another.

```python
def _tiktoken_encoder(encoding):
    if encoding in _TIKTOKEN_ENCODERS:
        return _TIKTOKEN_ENCODERS[encoding]
    try:
        import tiktoken
    except ImportError as exc:
        _TIKTOKEN_ENCODERS[encoding] = None
        return None                       # missing dep -> degrade
    try:
        encoder = tiktoken.get_encoding(encoding)
    except Exception as exc:
        raise ValueError(f"Invalid tiktoken encoding {encoding!r}: {exc}") from exc
    _TIKTOKEN_ENCODERS[encoding] = encoder
    return encoder
```

**2. Missing dependency vs. invalid encoding.** A missing `tiktoken` package still returns `None` so the caller degrades to the estimate counter with a warning (legitimate graceful fallback for an optional dependency). But an invalid configured encoding now **raises `ValueError`** instead of silently switching the counting unit — and invalid lookups are *not* cached, so they don't stick.

**3. Tests** (`TestContextBudgetTiktokenCache`, cache cleared per-test for isolation):
- `test_each_instance_uses_its_configured_encoding` — instances on `cl100k_base` and `p50k_base` each match their own encoder's token count for `"def __init__(self):"` (6 vs 7 tokens — so reusing the wrong encoder would miscount), and the cache holds two distinct encoder objects.
- `test_invalid_encoding_raises_configuration_error` — `not_a_real_encoding` raises `ValueError` mentioning the name.
- `test_invalid_encoding_before_valid_does_not_poison` — the core regression: an invalid encoding raises and is not cached, then a valid `cl100k_base` instance still counts real tokens.
- `test_missing_dependency_degrades_instead_of_raising` — with `tiktoken` import blocked, the processor degrades to the estimate rather than raising.

All 24 tests in the file pass.

hxbszyy (approved):
This fixes the configuration-order bug cleanly. Encoders are cached per encoding name, invalid configuration raises without poisoning later instances, and the missing-dependency fallback remains explicit. I also verified the two-encoding and invalid-before-valid cases in the focused suite: 24 tests pass. Approved.

PR #640, opened 2026-07-16
feat(knowledge): SensitiveDataRedactor for PII redaction (#248)

123123213weqw:
## What

Adds `SensitiveDataRedactor`, a knowledge *post-processor* (#248) that redacts common PII / sensitive identifiers from recalled documents **before** they reach the LLM, so personal data and secrets do not leak into model context.

## Why this component

It is distinct from every existing doc processor — none of them alter text for privacy/compliance. This fills that gap with a real, security-relevant capability (PII redaction is a standard RAG/guardrail feature) that is not inline-able: multiple high-precision patterns, configurable entities, custom patterns, and per-document summaries.

## Honest, high-precision detection

Regex-based, deterministic, dependency-free. Built-in entities are deliberately **high-precision** (structured formats, not loose guesses) to avoid noisy over-matching:

| entity | matches |
|---|---|
| `email` | standard email addresses |
| `credit_card` | 13–16 digit runs |
| `id_card` | China resident ID (18-digit, date-structured) |
| `ssn` | US SSN `ddd-dd-dddd` |
| `ip_address` | IPv4 |
| `api_key` | known prefixes `sk-` / `AKIA` / `ghp_` / `glpat-` |
| `phone` | opt-in (fuzzier, off by default) |

`custom_patterns` adds domain-specific identifiers; an invalid regex is skipped with a warning rather than crashing retrieval. Each match is replaced with a configurable `replacement`, and a per-document `redaction_summary` records the counts.

## Config

```yaml
name: 'sensitive_data_redactor'
metadata:
  type: 'DOC_PROCESSOR'
  module: 'agentuniverse.agent.action.knowledge.doc_processor.sensitive_data_redactor'
  class: 'SensitiveDataRedactor'
entities: [email, credit_card, id_card, ssn, ip_address, api_key]
replacement: '[REDACTED]'
custom_patterns: []
log_key: 'redaction_summary'
```

## Tests

`tests/.../test_sensitive_data_redactor.py` — 21, all pass:
- **Entities**: each built-in type is detected and replaced; `phone` is off by default and on when enabled; multiple entities in one doc.
- **Config**: custom replacement; custom pattern redaction; invalid regex / missing fields / unknown entity are skipped gracefully; `log_key: null` omits the summary; existing metadata is preserved; attributes load from the configer.
- **Integration**: redacts through the real `Knowledge.query_knowledge` → `post_processors` path.

`python -m unittest test_sensitive_data_redactor` → 21 passed.

Addresses the privacy / compliance direction of #248.

### DIFF
--- a/agentuniverse/agent/action/knowledge/doc_processor/sensitive_data_redactor.py
+++ b/agentuniverse/agent/action/knowledge/doc_processor/sensitive_data_redactor.py
@@ -14,13 +14,19 @@ issue #248 (knowledge post-processing components) and is distinct from every
 existing doc processor: none of them alter text for privacy/compliance.
 
 Detection is regex-based, deterministic, and dependency-free, so it is fully
-unit-testable offline. The built-in entity patterns are deliberately
-high-precision (structured formats such as emails, credit-card digit runs,
-China resident-id, US SSN, IPv4 addresses, well-known API-key prefixes) to
-avoid the noisy over-matching that a loose "find anything private" heuristic
-would produce. ``phone`` is available but opt-in, since phone matching is
-inherently fuzzier. Callers can add ``custom_patterns`` for domain-specific
-identifiers.
+unit-testable offline. The built-in entity patterns combine a structured
+*shape* (emails, credit-card digit runs, China resident-id, US SSN, IPv4
+addresses, well-known API-key prefixes) with a semantic check so a
+plausible-looking but invalid value is not redacted: credit cards pass the
+Luhn checksum, IPv4 octets are in range, China resident IDs and US SSNs pass
+their structural/checksum rules. This keeps false positives low without the
+noisy over-matching a loose "find anything private" heuristic would produce.
+``phone`` is available but opt-in, since phone matching is inherently fuzzier.
+Callers can add ``custom_patterns`` for domain-specific identifiers.
+
+Configuration is validated eagerly: an unknown entity, a malformed custom
+entry, or an invalid regex raises at load time rather than being silently
+skipped — a privacy component must not fail open.
 
 Each match is replaced with ``replacement`` (default ``[REDACTED]``); an
 optional per-document ``redaction_summary`` records how many of each entity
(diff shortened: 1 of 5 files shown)


hxbszyy (changes requested):
A pre-LLM redaction processor is directionally valuable, but a privacy component cannot fail open on configuration. Today an unknown entity name, a malformed custom entry, or an invalid regex is only logged and skipped, so a typo can silently disable the protection the operator believes is active. Please validate and compile the configuration at initialization and raise a clear error for any unknown or malformed entry.

The built-ins also need either semantic validation or a narrower contract. Several patterns described as “high precision” are shape checks only: any 13–16 digit run is treated as a card, invalid IPv4 octets match, and ID/SSN checks accept structurally invalid values. Please add validation such as Luhn/IP/date/checksum/range checks, or explicitly position and name this as best-effort pattern masking rather than a privacy/compliance guardrail. Cover both false positives and false negatives in tests.

123123213weqw:
Thanks @hxbszyy — both points addressed in 3bd4eedd.

**1. No more fail-open on configuration.** `_compiled_patterns` now raises `ValueError` on an unknown entity, a malformed `custom_patterns` entry (not a dict / missing `name` or `pattern` / empty or non-string), or an invalid regex. `_initialize_by_component_configer` validates eagerly, so a typo fails at **load time** rather than silently disabling redaction at processing time. (`process_docs` also raises, so direct construction can't bypass it.)

**2. Semantic validation, not just shape.** Built-in entities now pair their regex with a check; a match that fails it is left untouched and uncounted:
- `credit_card` → Luhn checksum
- `ip_address` → octets in 0–255
- `id_card` (China resident ID) → checksum digit
- `ssn` → valid area (not 000/666/900–999), group (not 00), serial (not 0000)

`email`/`api_key`/`phone` keep shape-only matching. The module docstring is updated to describe the shape+check contract.

**Tests.** The three tests that encoded the old log-and-skip behavior now expect `ValueError` (`test_unknown_entity_raises`, `test_invalid_custom_regex_raises`, `test_custom_pattern_missing_fields_raises`), plus a new `test_bad_config_raises_at_load_time` covering eager init validation. New `TestSensitiveDataSemanticValidation` covers both directions — valid values are redacted, invalid ones (non-Luhn `1234567890123456`, `999.999.999.999`/`1.2.3.256`, bad-checksum ID, and invalid SSN area/group/serial) are left in place with an empty summary.

All 26 tests pass.

hxbszyy (approved):
This now has the contract a privacy component needs: malformed configuration fails during initialization, and the high-risk structured patterns are backed by semantic validation rather than shape alone. The positive and negative cases are covered through the knowledge pipeline; 26 focused tests pass. Approved.

PR #646, opened 2026-07-17
fix(knowledge): add request timeout and non-JSON guard to JinaReranker

123123213weqw:
## Summary
`JinaReranker._process_docs` called `requests.post` without a timeout, so a stalled Jina rerank API hung the whole rerank step indefinitely (`requests` defaults to no timeout).

## Changes
- Add a configurable `request_timeout` attribute (default `30`s, overridable via yaml) and pass it to `requests.post`.
- Catch `ValueError` from `response.json()` so a non-JSON body (e.g. an HTML error page from an upstream gateway) surfaces a clear message instead of a raw parse error.
- `request_timeout` is read from the component configer like `model_name` / `top_n`.

## Tests
Updated the existing call-arg assertions (the call now carries `timeout`) and added coverage for the default/custom timeout, the timeout-error path, and the non-JSON path. `python -m unittest` — 9 passed.

### DIFF
--- a/agentuniverse/agent/action/knowledge/doc_processor/jina_reranker.py
+++ b/agentuniverse/agent/action/knowledge/doc_processor/jina_reranker.py
@@ -73,13 +73,18 @@ class JinaReranker(DocProcessor):
             response = requests.post(api_base, headers=headers, json=payload,
                                      timeout=self.request_timeout)
             response.raise_for_status()
-            results = response.json().get("results", [])
         except requests.exceptions.RequestException as e:
             raise Exception(f"Jina AI rerank API call error: {e}")
+
+        # Decode the body separately from the HTTP call. ``Response.json()``
+        # raises ``requests.exceptions.JSONDecodeError`` on a non-JSON body
+        # (e.g. an HTML error page from an upstream gateway). That exception
+        # inherits from both ``RequestException`` and ``ValueError``, so a
+        # single combined try block would misreport it as an API-call error;
+        # isolating JSON decoding here surfaces it as a non-JSON response.
+        try:
+            results = response.json().get("results", [])
         except ValueError as e:
-            # A non-JSON body (e.g. an HTML error page returned by an upstream
-            # gateway) makes response.json() raise ValueError; surface a clear
-            # message instead of letting the raw parse error escape.
             raise Exception(f"Jina AI rerank API returned a non-JSON response: {e}")
 
         rerank_docs = []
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
Adding a finite request timeout is valuable, but the non-JSON guard does not catch the real `requests` failure path as intended. `Response.json()` raises `requests.exceptions.JSONDecodeError`, whose MRO includes `RequestException`; because the `RequestException` handler comes first, an actual non-JSON response is reported as the generic API-call error. The test only raises plain `ValueError`, so it misses this behavior.

Please catch `requests.exceptions.JSONDecodeError` before `RequestException` (or isolate JSON decoding in its own try block), and exercise an actual `requests.Response` with a non-JSON body. Please also reject non-positive/non-numeric `request_timeout` values during initialization so bad configuration fails clearly.

123123213weqw:
Thanks @hxbszyy — all three points addressed in a83aafcf:

**1. Non-JSON guard ordering (MRO bug).** Confirmed `requests.exceptions.JSONDecodeError` inherits from *both* `RequestException` and `ValueError`, so the old combined `try` block caught `RequestException` first and misreported a real non-JSON body as a generic "API call error". I split JSON decoding into its own `try` block so it is surfaced as a non-JSON response.

```python
try:
    response = requests.post(...)
    response.raise_for_status()
except requests.exceptions.RequestException as e:
    raise Exception(f"Jina AI rerank API call error: {e}")

try:
    results = response.json().get("results", [])
except ValueError as e:
    raise Exception(f"Jina AI rerank API returned a non-JSON response: {e}")
```

**2. Test now exercises the real failure path.** It uses an actual `requests.Response` with an HTML body (`response._content = b'<html>...'`), which raises `requests.exceptions.JSONDecodeError` — the real exception — not a plain `ValueError`. This test fails on the old code (caught as "API call error") and passes on the new code, so it genuinely guards the regression. I also assert the message does *not* contain "API call error".

**3. `request_timeout` validated at init.** `_initialize_by_component_configer` now rejects non-numeric (`'not-a-number'`, `None`, `[30]`), non-positive (`0`, `-5`), and boolean (`True`/`False`, since `bool` is an `int` subclass and YAML `true` would silently become `1`) values, raising a clear `request_timeout` error. Covered by `test_initialize_rejects_non_numeric_request_timeout`, `test_initialize_rejects_non_positive_request_timeout`, and `test_initialize_accepts_valid_request_timeout`.

All 12 tests in `test_jina_reranker.py` pass.

hxbszyy (approved):
The real requests JSONDecodeError path is now isolated from RequestException handling, and timeout validation rejects unbounded or invalid configuration at load time. I reran the focused suite: all 12 tests pass. Approved.

PR #647, opened 2026-07-17
fix(reader): guard YuqueReader against non-JSON and malformed book payloads

123123213weqw:
## Summary
`YuqueReader` had three unguarded failure paths that crashed a whole-book read instead of degrading gracefully.

## Changes
- `_fetch_page_markdown` called `resp.json()` with no try/except, so a `200` response with a non-JSON body (e.g. an upstream gateway error page) raised `ValueError` out of the reader. Wrap it and return `''`.
- `_load_data` indexed `docs['book']['toc']` / `docs['book']['id']` outside its try block, so a decoded payload missing the `book` key (or carrying a non-dict value) raised `KeyError`. Use `.get` chains and return `[]` when the structure is unexpected.
- `__del__` used a bare `except`. Narrow to `except Exception`.

## Tests
Added unit tests (session stubbed) covering the non-JSON body, the non-200 status, and both malformed-book shapes. `python -m unittest` — 4 passed.

### DIFF
--- a/agentuniverse/agent/action/knowledge/reader/cloud_file_reader/yuque_reader.py
+++ b/agentuniverse/agent/action/knowledge/reader/cloud_file_reader/yuque_reader.py
@@ -10,6 +10,7 @@ import re
 import time
 import random
 import json
+import logging
 import urllib.parse
 import requests
 from pydantic import ConfigDict
@@ -19,6 +20,8 @@ from typing import List, Any, Optional
 
 from agentuniverse.agent.action.knowledge.reader.reader import Reader
 from agentuniverse.agent.action.knowledge.store.document import Document
+
+logger = logging.getLogger(__name__)
 
 
 class YuqueReader(Reader):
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
The added guards avoid a crash, but they turn malformed/auth/error responses into successful-looking empty ingestion. A missing/non-dict `book`, invalid page JSON, missing book id, or missing item URL should be observable to the caller; returning `[]` / `""` and printing to stdout makes data loss very hard to diagnose. The page guard is also incomplete because a valid JSON list/string still fails at `.get()`.

Please validate the complete response shape and surface a clear reader error (or at minimum log it through the project logger with the URL/book context) rather than silently treating it as no content. Add cases for non-dict page JSON and missing required ids/URLs. This should also be rebased with the active reader consolidation work so we do not harden only a path that is being moved.

123123213weqw:
Thanks @hxbszyy — the robustness issues are addressed in c1b96cfa. On the rebase, I want to flag a judgment call rather than do it silently.

**Robustness (done):**
- Replaced every `print` with a module `logger`; failures now log with `book_id`/`slug`/`url` context.
- `_load_data` validates the full decoded shape and **raises `RuntimeError`** (with the URL) when the payload is not an object, has no `book` object, or the book has no `id` — so a deleted / private / auth-failed book is observable instead of a silent `[]`.
- `_fetch_page_markdown` now checks the JSON body is an object *before* reading `data`, so a valid JSON list/string (the gap you flagged — it used to crash at `.get('data')`) is logged and skipped. Non-dict `data` is handled too. `toc` entries missing `url` are logged and skipped rather than passed through as `None`.
- Tests: the two book-level cases now expect `RuntimeError`; added non-dict page JSON, missing book id, non-dict payload, toc-entry-missing-url, a page happy path, and an end-to-end load; log context is asserted. 10 tests pass.

**On rebasing onto the reader consolidation (#634):** I checked and #634 *deletes* `cloud_file_reader/yuque_reader.py` (the file this PR hardens) and replaces it with a new `cloud/yuque_reader.py` that already adopts `logging` + the typed `reader_errors` hierarchy. I deliberately did **not** rebase onto #634, because:
- #634 opened yesterday and is still being reviewed by another author — rebasing now couples this PR to an evolving branch and would bloat the diff with unrelated restructuring.
- If #634 is reworked or rejected, a #647 built on top of it breaks.

My suggestion: land this hardening on the current path now (it's correct and tested), then **carry these exact guards into `cloud/yuque_reader.py` when #634 merges** — and worth noting, #634's new `_fetch_page_markdown` still does `resp.json().get("data", {})`, so it has the same non-dict-JSON gap and should pick up the object check during that migration.

If you'd prefer I rebase onto #634 now and apply the fixes directly to `cloud/yuque_reader.py` (accepting the cross-PR dependency), say the word and I'll do it.

hxbszyy (approved):
The current-path hardening stands on its own, so I agree that coupling it to the still-unaccepted #634 refactor would be the wrong dependency. Book-level authentication and shape failures are now observable, page-level skips carry structured diagnostics, and the missing URL and non-dict JSON gaps are covered. All 10 focused tests pass. Please carry these guards forward if the reader layout is later consolidated. Approved.

PR #648, opened 2026-07-17
chore(reader): remove leftover debug print statements from WebPageReader

123123213weqw:
## Summary
`WebPageReader` shipped with twelve `print(f"debugging: ...")` statements across `_load_data`, `_fetch_html` and `_extract_main_text` that pollute the host process's stdout on every use of the reader.

## Changes
- Drop all twelve debug prints. No behavior changes.
- The three extractor-fallback handlers that only printed now use a plain `except Exception: pass` / `raise`, and the now-unused exception bindings are removed.

No test added: pure stdout cleanup with no behavioral change. Verified with `python -m py_compile`.

- replaced stdout prints with a module logger and debug-level fallback diagnostics including exception context
- added focused tests for clean stdout, httpx fallback logging, extractor logging, and terminal failure chaining
- implementation is in `49bc1593`

### DIFF
--- a/agentuniverse/agent/action/knowledge/reader/web/web_page_reader.py
+++ b/agentuniverse/agent/action/knowledge/reader/web/web_page_reader.py
@@ -24,15 +24,11 @@ class WebPageReader(Reader):
     """
 
     def _load_data(self, url: str, ext_info: Optional[Dict] = None) -> List[Document]:
-        print(f"debugging: WebPageReader start load url={url}")
         if not isinstance(url, str) or not url:
             raise ValueError("WebPageReader._load_data requires a non-empty url string")
 
         html = self._fetch_html(url)
-        print(f"debugging: WebPageReader fetched html length={len(html)}")
-
         text, metadata_extra = self._extract_main_text(html, url)
-        print(f"debugging: WebPageReader extracted text length={len(text)}")
 
         metadata: Dict = {"source": "web", "url": url}
         metadata.update(metadata_extra)
@@ -44,7 +40,6 @@ class WebPageReader(Reader):
     def _fetch_html(self, url: str) -> str:
         try:
             import httpx  # type: ignore
-            print("debugging: WebPageReader using httpx")
             with httpx.Client(timeout=20.0, headers={
                 "User-Agent": "agentUniverse/1.0 (+https://github.com/)",
                 "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
@@ -53,10 +48,8 @@ class WebPageReader(Reader):
                 resp.raise_for_status()
                 return resp.text
         except Exception as e_httpx:
-            print(f"debugging: WebPageReader httpx failed: {e_httpx}")
             try:
                 import requests  # type: ignore
-                print("debugging: WebPageReader using requests fallback")
                 resp = requests.get(url, timeout=20, headers={
                     "User-Agent": "agentUniverse/1.0 (+https://github.com/)",
                     "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
Removing unconditional debug prints from a library reader is the right goal, but replacing the extractor and fallback failures with `except Exception: pass` removes the only diagnostics. When httpx, trafilatura, or readability fails, operators now cannot tell whether a fallback was selected normally or because of a dependency, parse, TLS, or response problem.

Please add a module logger and record fallback failures at an appropriate debug/warning level (including exception context) while keeping stdout clean. The final extractor failure can still raise as it does today. A focused test should also verify that normal reads do not write to stdout.

123123213weqw:
Thanks @hxbszyy — addressed in 49bc1593.

**1. Diagnostics go through a module logger.** Added `logger = logging.getLogger(__name__)`. Each extractor/fetch fallback that previously did `except Exception: pass` now logs at DEBUG with `exc_info=True` and the originating URL:

```python
except Exception as e:
    logger.debug("trafilatura extraction failed for %s, trying next extractor: %s",
                 url, e, exc_info=True)
```

This lets operators tell whether a fallback was selected normally (empty result is logged separately) or because of a missing dependency / parse / TLS / response error — while stdout stays clean. The httpx→requests fetch fallback is logged the same way.

**2. Final extractor still raises.** The bs4 branch still raises `RuntimeError` (now chained `from e` to preserve the root cause).

**3. Tests.** New `test_web_page_reader.py`:
- `test_normal_read_does_not_write_to_stdout` — patches `_fetch_html`, runs a full read through `redirect_stdout`/`redirect_stderr`, and asserts both stay empty while a Document is still produced.
- `test_fallback_failures_are_logged_not_swallowed` — in this env trafilatura/readability are uninstalled, so they ImportError; the test asserts the fallback is logged (mentioning the URL) with `exc_info` set, and bs4 still produces the document.
- `test_missing_all_extractors_raises` — blocks all three extractor imports and asserts `RuntimeError` rather than silent empty content.

All 3 tests pass.

123123213weqw:
Friendly re-review ping @hxbszyy — both points from your review were addressed in 49bc1593 (7 days ago):

1. **Module logger with exception context.** Each extractor/fetch fallback that previously did `except Exception: pass` now logs at DEBUG with `exc_info=True` and the originating URL (e.g. `logger.debug("trafilatura extraction failed for %s, trying next extractor: %s", url, e, exc_info=True)`). Normal empty-result fallbacks are logged separately so an operator can tell "no content" from "dependency/parse/TLS error".
2. **Final extractor still raises.** The bs4 branch raises `RuntimeError` (chained `from e`).
3. **Stdout stays clean — tested.** `test_normal_read_does_not_write_to_stdout` patches `_fetch_html` and asserts both stdout and stderr stay empty while a Document is still produced. The two companion tests cover fallback-logged-with-exc_info and missing-all-extractors-raises.

Status is `MERGEABLE` / `CLEAN`, all 3 tests pass. Ready when you are.

hxbszyy (changes requested):
The logging change itself is in good shape, but the new regression file is not portable across supported reader installations. I installed the optional `trafilatura` dependency and reran it: 2 of 3 tests fail. The fallback test assumes trafilatura/readability are absent and insists on the bs4 extractor; the “normal read” test also assumes bs4 removes `noscript`, while trafilatura legitimately returns that text.

Please make the tests control the branch they are testing. Mock the trafilatura/readability imports or extraction calls to raise for the fallback-log case, and keep the stdout test focused on stdout/stderr plus successful document production rather than extractor-specific output. The suite should pass both in a minimal install and with the reader extras installed. Once that is fixed, the implementation is ready.

123123213weqw:
Thanks @hxbszyy — addressed in accb46c2. Each test now controls the branch under test directly, so the suite is portable across installations.

**The fix, point by point:**

1. **Fallback test no longer assumes deps are absent.** `trafilatura` and `readability` are now injected as fakes via a `_ModuleGuard` context manager that swaps them into `sys.modules` (and restores prior state on exit). The fallback test forces `trafilatura.extract` to raise `RuntimeError` and `readability.Document.__init__` to raise, so the bs4 branch is reached deterministically regardless of what is installed. Both failures are still asserted to be logged with the URL + `exc_info`.

2. **Normal-read test no longer asserts extractor-specific output.** It now injects a stub `trafilatura` that returns content, then asserts only: (a) one document is produced, (b) stdout/stderr stay clean. No claim about `noscript` removal or which extractor won.

3. **New empty-result fall-through test** covers the path where trafilatura succeeds but returns empty content — asserts a "returned no content" debug log (not an exception log) and that the bs4 fallback still produces a document.

**Validation:** I reproduced the original failure by `pip install trafilatura readability-lxml` — the old tests went 2/3 red. After this fix the suite is 4/4 green in both environments:
- minimal install (deps absent → ImportError): `Ran 4 tests in 0.052s OK`
- reader extras installed (trafilatura 2.1.0, readability-lxml 0.8.4.1): `Ran 4 tests in 0.708s OK`

The implementation is unchanged from 49bc1593; only the test file was rewritten.

PR #650, opened 2026-07-17
fix(agent): avoid mutating configured tool names

messere1:
## Summary

Copy the configured direct-tool list before appending toolkit tools so reading `Agent.tool_names` remains side-effect free.

Before this change, each property access permanently appended toolkit entries to `AgentModel.action["tool"]`; repeated agent runs could therefore invoke the same toolkit tool multiple times.

## Tests

- `python -m pytest tests/test_agentuniverse/unit/agent/test_agent.py -k "tool_names or process_prompt" -q` (2 passed)
- `ruff check --no-fix tests/test_agentuniverse/unit/agent/test_agent.py`
- `python -m black --check tests/test_agentuniverse/unit/agent/test_agent.py`
- `python -m py_compile agentuniverse/agent/agent.py tests/test_agentuniverse/unit/agent/test_agent.py`

### DIFF
--- a/agentuniverse/agent/agent.py
+++ b/agentuniverse/agent/agent.py
@@ -377,9 +377,7 @@ class Agent(ComponentBase, ABC):
         return self._get_tool_names()
 
     def _get_tool_names(self) -> list:
-        tool_name_list = self.agent_model.action.get('tool', [])
-        if tool_name_list is None:
-            tool_name_list = []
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
This fixes a real state leak with a minimal change. Reading tool_names no longer mutates AgentModel.action, and repeated reads remain stable. I reran the agent tests in the project environment: 2 tests pass. Approved.

PR #651, opened 2026-07-17
fix(agent): handle empty streaming responses

messere1:
## Summary

Return an empty string when a streaming agent chain produces no chunks instead of indexing an empty list and raising `IndexError`.

This preserves the existing result type for text streams and lets callers handle a legitimate empty model response without an unrelated aggregation failure.

## Tests

- `python -m pytest tests/test_agentuniverse/unit/agent/test_agent.py -q` (2 passed)
- `ruff check --no-fix tests/test_agentuniverse/unit/agent/test_agent.py`
- `python -m black --check tests/test_agentuniverse/unit/agent/test_agent.py`
- `python -m py_compile agentuniverse/agent/agent.py tests/test_agentuniverse/unit/agent/test_agent.py`

### DIFF
--- a/agentuniverse/agent/agent.py
+++ b/agentuniverse/agent/agent.py
@@ -363,6 +363,8 @@ class Agent(ComponentBase, ABC):
         return streaming
 
     def generate_result(self, data: list[dict | str]):
+        if not data:
+            return ""
         if isinstance(data[0], str):
             return "".join(data)
         text = [val.get('text') for val in data]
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
An empty stream should not turn into an unrelated IndexError during aggregation. Returning the existing text result type as an empty string is a small, compatible fix, and the focused agent tests pass. Approved.

PR #658, opened 2026-07-17
fix(tool): verify TLS on Jina API requests by default

123123213weqw:
## Summary

`JinaAITool._make_api_request` called `requests.get(url, verify=False, ...)`, **disabling TLS certificate verification on every Jina read / search / fact-check request**. A network attacker could man-in-the-middle the connection and inject or alter the web/search content that flows into the agent's context (a prompt-injection vector over the wire). Jina's API serves a valid certificate, so there is no reason to disable verification.

## Changes

- Add a `verify_tls: bool = True` field so requests verify certificates by default (secure), while remaining configurable for a legitimate trusted self-signed endpoint.
- Replace a stray `print()` of a failed response code with `logger.warning()` (keeps stdout clean, consistent with the reader/logger work in #648/#647).
- Add `tests/test_jina_ai_tool.py` covering the secure default, the opt-out, and that a non-200 response is reported via the logger (stdout stays clean).

## Test plan

```
python -m unittest tests.test_agentuniverse.unit.agent.action.tool.test_jina_ai_tool -v
```
All 3 tests pass.

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/jina_ai_tool.py
+++ b/agentuniverse/agent/action/tool/common_tool/jina_ai_tool.py
@@ -11,6 +11,7 @@ import time
 from pydantic import Field
 from agentuniverse.agent.action.tool.tool import Tool, ToolInput
 from agentuniverse.base.util.env_util import get_from_env
+from loguru import logger
 import html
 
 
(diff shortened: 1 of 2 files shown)


123123213weqw:
Resolved the add/add conflict against the latest `master`. The combined test file now covers both the TLS-verification behavior from this PR and the upstream regression for `HTTPError` instances without a response.

Validation: `python -m unittest tests.test_agentuniverse.unit.agent.action.tool.test_jina_ai_tool` — 4 tests passed; Ruff check/format passed for the reconciled test file.

hxbszyy (approved):
TLS verification should be the default here, with an explicit escape hatch for trusted self-signed endpoints. The request behavior and regression coverage look sound; I reran the focused suite (4 tests). Approved.

PR #659, opened 2026-07-17
fix(store): persist FAISS metadata as JSON instead of pickle

123123213weqw:
## Summary

`FAISSStore._load_index_and_metadata` loaded document metadata with `pickle.load`. **Pickle deserialization of a writable `metadata_path` file is an arbitrary-code-execution vector** — a malicious or tampered metadata file runs code on load. The path is taken from component configuration, so any file the process can read is in scope (shared filesystems, config injection, supply-chain).

Persist the metadata as **JSON** instead, which cannot execute code on load.

## Changes

- `_read_metadata_file` / `_write_metadata_file` factor the metadata I/O into testable helpers. Load uses `json.load`; save uses `json.dump`.
- `Document` objects serialize via `model_dump(mode="json")` and restore via the `Document` constructor; `Set` fields (keywords) round-trip correctly.
- Integer FAISS positions (`index_to_id` keys) survive the JSON string-key round-trip (`int(k)` on load).
- A `format: "faiss-store-metadata-v1"` tag is written for future compatibility.
- **Backward compatibility / migration:** a legacy pickle-format file is no longer loadable. The loader reports it as invalid JSON with a clear message explaining the deprecation (pickle = RCE vector) and that the store should be re-indexed, then resets to empty rather than deserializing it. This is intentional — auto-loading legacy pickle would reintroduce the vulnerability.

## Tests

Added `TestFAISSStoreMetadataSerialization`, which runs **without** the optional FAISS dependency (it exercises the metadata helpers directly):
- `test_metadata_round_trips_as_json` — text / metadata / embedding / keywords / integer index keys all survive a save→load cycle, and the file on disk is genuine JSON.
- `test_missing_metadata_file_returns_false`.
- `test_legacy_pickle_payload_is_not_executed` — writes an RCE pickle payload (`__reduce__` → `open(marker)`); asserts the loader returns false, resets to empty, and **never executes the payload** (marker file absent).

Locally: `17 tests, 3 passed, 14 skipped` (the 14 existing FAISS tests skip without `faiss` installed and run in CI).

```
python -m unittest tests.test_agentuniverse.unit.agent.action.knowledge.store.test_faiss_store
```

- validate the format marker and complete JSON payload shape, including documents and integer index keys
- reset FAISS index plus metadata coherently for every reconstruction failure
- write metadata through a same-directory temporary file, `fsync`, and atomic `os.replace`
- fixed the module-level FAISS import lifecycle exposed by a real `faiss-cpu` installation
- added valid-but-corrupt JSON, unsupported-version, reconstruction-failure, atomic-write interruption, and full installed-FAISS regressions
- implementation is in `10d143a9`

### DIFF
--- a/agentuniverse/agent/action/knowledge/store/faiss_store.py
+++ b/agentuniverse/agent/action/knowledge/store/faiss_store.py
@@ -8,14 +8,30 @@
 import json
 import logging
 import os
+import tempfile
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any
 
 from agentuniverse.agent.action.knowledge.embedding.embedding_manager import EmbeddingManager
 from agentuniverse.agent.action.knowledge.store.document import Document
 from agentuniverse.agent.action.knowledge.store.query import Query
 from agentuniverse.agent.action.knowledge.store.store import Store
 from agentuniverse.base.config.component_configer.component_configer import ComponentConfiger
+
+# faiss and numpy are imported at module scope so every method (query / insert /
+# _create_faiss_index / ...) sees the same names. They are optional dependencies:
+# a hard module-level import would break environments without faiss installed,
+# so each is wrapped and ``_new_client`` surfaces a clear ImportError when the
+# store is actually used.
+try:
+    import faiss  # type: ignore[import-not-found]
+except ImportError:  # pragma: no cover - exercised via the skip guard in tests
+    faiss = None  # type: ignore[assignment]
+
+try:
+    import numpy as np  # type: ignore[import-not-found]
+except ImportError:  # pragma: no cover - numpy ships with faiss-cpu
+    np = None  # type: ignore[assignment]
 
 # Default configuration for FAISS index types
 DEFAULT_INDEX_CONFIG = {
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
The security goal is right, but this on-disk format is not robust enough to merge yet. The `format` marker is written and never validated, and a syntactically valid JSON file with the wrong shape, invalid `Document` data, or non-integer `index_to_id` keys escapes `_read_metadata_file()` instead of resetting cleanly. That can leave the loaded FAISS index and metadata in an incoherent state. The write is also in-place, so an interrupted write destroys the last usable metadata file.

Please validate the version and full payload schema, recover/reset the index and metadata as one coherent unit for every reconstruction failure, and write through a temporary file plus atomic replace. Add corrupt-but-valid-JSON and interrupted-write regressions. I also installed `faiss-cpu` and reran the whole test file: the three new serialization tests pass, but the previously skipped FAISS suite exposes 11 failures, mostly because `faiss` is only imported locally in `_new_client()` while the rest of the module expects a module-level name. Please run and fix the suite with FAISS actually installed rather than relying on the skip path.

123123213weqw:
Thanks @hxbszyy — every point addressed in 10d143a9.

**1. Module-scope faiss/numpy imports (the 11 failing tests).** Both optional deps were only imported inside `_new_client`, so `query` / `insert_document` / `_create_faiss_index` referenced undefined names once the suite stopped being skipped. They are now imported once at module scope (wrapped so a missing faiss still surfaces a clear `ImportError` from `_new_client`), and the whole FAISS suite runs instead of silently skipping.

**2. Schema validation.** The `format` marker is now checked, and the full payload — top-level shape, each `document_store` entry, `id_to_index`, `index_to_id` integer keys, `next_index` — is validated in `_validate_metadata_payload` *before* any field is assigned. A corrupt-but-valid-JSON file (wrong format marker, non-object top level, malformed `Document` payload, or non-integer `index_to_id` key) resets the store to empty as **one coherent unit**, so the loaded FAISS index and metadata can never end up inconsistent.

**3. Atomic writes.** `_write_metadata_file` writes through a sibling temp file, `fsync`s, and atomically `os.replace`s the target, so an interrupted write cannot destroy the last usable metadata file. Stale temp files are cleaned up on failure.

**Regressions added** (5):
- `test_wrong_format_marker_resets_store`
- `test_top_level_non_object_resets_store`
- `test_invalid_document_payload_resets_store`
- `test_non_integer_index_to_id_key_resets_store`
- `test_metadata_write_is_atomic` — patches `os.replace` to fail mid-write, asserts the previous good file is intact and no temp file leaked.

Validation: `pip install faiss-cpu` then `python -m unittest tests.test_agentuniverse.unit.agent.action.knowledge.store.test_faiss_store` — **22 tests pass** (17 prior + 5 new), no skips; `ruff check` / `ruff format` clean (ruff errors on this file dropped from 36 → 8, all remaining are TRY003 style warnings consistent with sibling stores like `milvus_store.py`).

hxbszyy (changes requested):
Thanks for fixing the imports, version marker, and atomic JSON write. All 22 tests now run and pass with FAISS installed. One end-to-end coherence bug remains, though, and it is serious.

`_load_index_and_metadata()` loads the FAISS index first. If metadata validation then fails, `_reset_metadata()` clears only the maps and leaves the loaded index intact. I reproduced this with one persisted document, a corrupted format marker, reload, and a new insert: after reload the store had `faiss_index.ntotal == 1`, zero documents, and `_next_index == 0`; after inserting the new document it had two vectors but only `{0: "new"}` in `index_to_id`. Querying the old vector then returned the new document. That is exactly the index/metadata incoherence the previous review asked us to prevent.

Please fail closed as one store: when either half cannot be loaded and validated, discard/recreate the in-memory index as well as all metadata (or raise and require re-indexing). Add an end-to-end regression through `_new_client()` with a valid persisted index plus corrupt metadata, followed by insert/query.

The payload validation also needs relational invariants, not only field types: `id_to_index` and `index_to_id` must be inverses, every ID must exist in `document_store` and match `Document.id`, positions and `next_index` must be non-negative/consistent, and the mapping cardinality/positions must agree with `faiss_index.ntotal`. Otherwise a typed but internally inconsistent JSON file still passes. Detecting that mismatch is especially important because the index and metadata are two separately replaced files and a crash can leave generations out of sync.

123123213weqw:
Thanks @hxbszyy — both points addressed in 1aaa89a5.

**1. Fail closed as one unit.** `_load_index_and_metadata` previously loaded the FAISS index first; if metadata validation then failed, `_reset_metadata` cleared the maps but left the loaded index in memory. I reproduced the exact regression you described (one persisted doc + corrupted format marker + reload → `ntotal=1`, zero documents, `_next_index=0`; insert → two vectors but only `{0: "new"}` mapped; query of the old vector returned the new document). The store now fails closed: when either half cannot be loaded and validated, both the in-memory index and all metadata are discarded, and a log line points the operator at re-indexing. The loaded index's `ntotal` is also threaded into metadata validation so a generation mismatch (right shape, wrong vector count) is caught rather than silently accepted.

**2. Relational invariants, not only field types.** `_validate_metadata_payload` now takes `faiss_ntotal` and, with a new `_check_mapping_invariants` helper, enforces:
- `id_to_index` and `index_to_id` are exact inverses;
- every mapped id exists in `document_store` and matches `Document.id`;
- positions and `next_index` are non-negative, and `next_index` is strictly greater than the highest used position (so the next insert cannot collide);
- when `faiss_ntotal` is supplied, the number of mapped positions agrees with the number of vectors the index actually holds (this is what makes the index/metadata generation mismatch visible).

**Regressions added (10):**
- *Invariant branches (7):* document id mismatch with key, non-inverse mappings, orphan id reference, `next_index` not past highest position, negative position, ntotal mismatch rejected, ntotal match accepted.
- *End-to-end through `_new_client()` with a real FAISS index (3):* corrupt-format reload discards the index; corrupt-reload + new insert never orphans a vector and querying never returns a stale mapping; generation mismatch between index (ntotal=1) and metadata (0 vectors) discards both.

Validation: `pip install faiss-cpu` then `python -m unittest tests.test_agentuniverse.unit.agent.action.knowledge.store.test_faiss_store` — **32 tests pass** (22 prior + 10 new), 0 skips.

Issue #259, opened 2025-01-16
Enhancement for VectorDB components/VectorDB新增


LandJerry:
======================== English version ==============================

期待您优先接入的VectorDB组件如下：
* FAISS
* 其他项目中未包含的同类组件

======================== 中文版 ==============================

The VectorDB components you are looking forward to prioritizing for access are as follows:

* FAISS
* Other VectorDB components not included in the current projects

VKttZXvC:
@LandJerry, i want to have a try, plz assign to me, thanks.

AniviaTn:
Thank you for your interest in contributing! We are very excited to have you on board. We hope that you will be able to provide the following for the VectorDB components:

Basic configuration for VectorDB
Knowledge entry based on Document from agentUniverse and knowledge query based on Query
Example usage for the corresponding VectorDB components
Please feel free to refer to the existing VectorDB sections in the [XXX documentation] and the practical examples in [YYY documentation] to help guide your commit.

Saswatsusmoy:
Hii @AniviaTn I was looking forward to taking up this issue. Answering your questions


**Implementation Plan:**

**Basic Configuration**: Complete FAISS store with YAML config support for different index types (IndexFlatL2, IndexIVFFlat, IndexHNSWFlat) and configurable parameters

**Knowledge Entry & Query**: Full CRUD operations with seamless integration of Document/Query classes, automatic embedding generation, and efficient similarity search

**Example Usage**: Multiple YAML configs, Python examples, and integration patterns with existing workflows

**Documentation**: Complete usage guide, installation instructions, and troubleshooting

**Additional Features**: Multiple index types, persistent storage, scalability optimization, error handling, and unit tests.

PR #661, opened 2026-07-17
feat(store): add PostgreSQL pgvector backend (#259)

123123213weqw:
## What

Adds `PGVectorStore`, a PostgreSQL + pgvector knowledge store for #259.

## Capabilities

- synchronous and asynchronous insert/upsert/update/delete/query;
- cosine, L2, and inner-product distance;
- JSONB metadata containment filters;
- precomputed embeddings or automatic embedding through `EmbeddingManager`;
- strict vector type/dimension validation with first-write dimension inference;
- automatic `vector` extension/table creation and optional HNSW index;
- configurable table, dimensions, distance, top-k, and connection URL;
- connection URL can be provided through `PGVECTOR_CONNECTION_URL` instead of committed configuration.

Document values and metadata filters are parameterized. Dynamic SQL contains only a table identifier validated against a strict identifier regex and an operator selected from a fixed allowlist.

## Dependency and configuration

`psycopg[binary]` and `pgvector` are optional members of the existing `store_ext` extra. A copyable `pgvector_store.yaml.example` is included rather than an auto-loaded live configuration, so importing/scanning the framework never attempts an unsolicited database connection.

English and Chinese Store guides document setup and credential handling.

## Tests

`python -m unittest test_pgvector_store` → **17 passed**.

The deterministic suite uses fake connections and covers extension/adapter initialization order, SQL/schema generation, table-name injection rejection, dimensions and top-k validation, metadata filtering, managed embeddings, parameterized upsert/delete, native `pgvector.Vector` row conversion, optional-dependency errors, environment configuration, and the real component metadata contract. In addition, the actual store passed synchronous and asynchronous CRUD/query/filter tests on Ubuntu 22.04 x86_64 with Python 3.10.12 against a Docker-hosted PostgreSQL 16 server with the pgvector extension and an HNSW index.

## Ubuntu server experiment

Repeated independently on an Ubuntu 22.04 x86_64 server with Python 3.10.12. A disposable `pgvector/pgvector:pg16` Docker container provided a real PostgreSQL 16 database; the container and test table were removed after the run.

- Ran all 17 deterministic unit and component-contract tests: all passed.
- Exercised the actual `PGVectorStore`, not a SQL-only surrogate, with `psycopg` 3.3.4 and `pgvector` 0.5.0.
- Verified extension-before-adapter initialization, table creation, three-dimensional vectors, HNSW index creation, cosine and L2 queries, JSONB containment filters, upsert/update/delete, native `pgvector.Vector` decoding, and both sync and async clients.
- Result: the complete sync/async CRUD-query-filter-index workflow passed against the real database.

Partially addresses #259 (additional VectorDB components).

### DIFF
--- a/agentuniverse/agent/action/knowledge/store/pgvector_store.py
+++ b/agentuniverse/agent/action/knowledge/store/pgvector_store.py
@@ -1,0 +1,272 @@
+#!/usr/bin/env python3
+"""PostgreSQL pgvector knowledge store."""
+
+# Dynamic SQL interpolates only table/operator identifiers validated against fixed allowlists.
+# ruff: noqa: TRY003, S608
+
+import json
+import os
+import re
+from typing import Any, ClassVar
+
+from agentuniverse.agent.action.knowledge.embedding.embedding_manager import EmbeddingManager
+from agentuniverse.agent.action.knowledge.store.document import Document
+from agentuniverse.agent.action.knowledge.store.query import Query
+from agentuniverse.agent.action.knowledge.store.store import Store
+from agentuniverse.base.config.component_configer.component_configer import ComponentConfiger
+
+
+class PGVectorStore(Store):
+    """Persistent vector store backed by PostgreSQL and the pgvector extension."""
+
+    connection_url: str | None = None
+    table_name: str = "agentuniverse_documents"
+    embedding_model: str | None = None
+    dimensions: int | None = None
+    distance: str = "cosine"
+    similarity_top_k: int = 10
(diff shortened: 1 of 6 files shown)


hxbszyy (approved):
This fills a real gap in the Store layer and stays within the existing component model. The identifier/operator boundaries, parameterized values, optional dependency handling, sync/async parity, and real PostgreSQL/pgvector validation make this more than a surface adapter. I reran all 17 focused tests. Approved.

PR #662, opened 2026-07-17
feat(tool): add bounded PDF manipulation tool (#252)

123123213weqw:
## What

Adds a built-in `PDFTool` under the additional tools direction of #252.

It provides five bounded modes:

- `merge` — combine PDFs and optionally set metadata;
- `split` — emit selected pages as individual PDFs;
- `rotate` — rotate selected pages into a new PDF;
- `extract` — return text from selected pages under a context budget;
- `info` — return page count, size, and metadata.

## Safety and behavior

- every source/destination directory is confined to `base_dir` by the shared safe-path resolver;
- validates PDF extension, `%PDF-` signature, encryption state, page ranges, rotation, and configuration;
- bounds input files, pages, read/write bytes, and extracted text;
- encrypted PDFs fail explicitly rather than producing partial output;
- existing outputs are never replaced without `overwrite=true`;
- all writes use a same-directory temporary file and atomic `os.replace`;
- `pypdf` is an optional lazy dependency exposed through `pdf_ext`.

## Integration

Includes a registerable built-in YAML, common-tool export, English/Chinese documentation, and registration tests through the real framework loader and `ToolManager`.

## Tests

`python -m unittest test_pdf_tool` → **15 passed**.

The suite creates real temporary PDFs offline and covers info, merge, atomic split-destination preflight, rotate, extraction, overwrite protection, page/rotation validation, safe paths, signature/page limits, missing dependency guidance, and component registration. A supplemental seeded stress run completed 75 merge/rotate/split/info workflows and verified every generated page. The full targeted suite and stress run were repeated successfully on Ubuntu 22.04 x86_64 with Python 3.10.12 in a clean virtual environment.

## Ubuntu server experiment

Repeated independently on an Ubuntu 22.04 x86_64 server with Python 3.10.12 in a clean virtual environment using `pypdf` 6.14.2.

- Ran all 15 targeted unit and framework-registration tests: all passed.
- Ran 75 deterministic randomized workflows (seed `20260718`) with 1–5 inputs and 1–8 pages per input.
- Each workflow merged PDFs, rotated a random page subset, split selected pages, reopened every output, and verified page counts and rotation values.
- Result: 75/75 workflows passed; every split output reopened as a valid one-page PDF.

Partially addresses #252 (additional/Office document tools).

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/__init__.py
+++ b/agentuniverse/agent/action/tool/common_tool/__init__.py
@@ -7,9 +7,11 @@
 # @FileName: __init__.py
 
 from .github_tool import GitHubTool
+from .pdf_tool import PDFTool
 from .yahoo_finance_tool import YahooFinanceTool
 
 __all__ = [
     'GitHubTool',
+    'PDFTool',
     'YahooFinanceTool',
 ]
--- a/agentuniverse/agent/action/tool/common_tool/pdf_tool.py
+++ b/agentuniverse/agent/action/tool/common_tool/pdf_tool.py
@@ -183,14 +183,15 @@ class PDFTool(Tool):
         _, Writer = self._classes()
         os.makedirs(directory, exist_ok=True)
         stem = os.path.splitext(os.path.basename(source))[0]
-        outputs = []
-        for index in indexes:
-            destination = os.path.join(directory, f"{stem}-page-{index + 1}.pdf")
+        # Preflight every destination before writing any page. This prevents a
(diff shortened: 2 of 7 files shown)


hxbszyy (approved):
This is a useful, deliberately bounded PDF workflow rather than a thin wrapper. Safe path resolution, read/write/page limits, encrypted-file rejection, atomic writes, and split preflight cover the failure modes I would expect here. I installed pypdf and reran the focused suite (15 tests); it passes. Approved.

Issue #500, opened 2025-10-30
Enhancement for agent context/智能体上下文贡献 #257


LandJerry:
======================== English version ==============================

Context engineering is a critical core capability in the current field of intelligent agents, directly determining an agent's depth of task understanding, coherence in dialogue, and effectiveness in executing complex tasks. We aim to benchmark against leading industry projects (such as Cursor, Claude, etc.) to build a comprehensive and efficient context management module and related processing mechanisms.

This project seeks to overcome these bottlenecks by establishing a robust context engineering system, enabling intelligent agents to handle more complex multi-turn tasks, possess long-term memory and dynamic knowledge updating capabilities, and intelligently perform context compression and refinement. It will also provide adaptive context management strategies for different types of tasks (such as code generation, data analysis, and dialogue systems).

======================== 中文版 ==============================

上下文工程是当前智能体领域至关重要的核心能力，它直接决定了智能体对任务的理解深度、对话的连贯性以及执行复杂任务的有效性。我们希望对标业内前沿的项目（如 Cursor、Claude等），构建一个完善、高效的上下文管理模块与相关处理机制。

本项目旨在通过建立一个强大的上下文工程体系，来突破这些瓶颈，使智能体能够处理更复杂的多轮任务，具备长期记忆和动态知识更新能力，更智能地进行上下文压缩与精炼，为不同类型的任务（如代码生成、数据分析、对话系统）提供自适应的上下文管理策略。

SeasonPilot:
Please take a look at whether this plan is feasible. If there are no issues, I will start implementing it.
@LandJerry 


## Technical Analysis

### Core Challenges Identified

1. **Context Management at Scale**
   - Multi-turn conversation coherence
   - Long-term memory persistence
   - Dynamic knowledge graph updates

2. **Context Optimization**
   - Intelligent compression algorithms
   - Selective refinement strategies
   - Performance vs. accuracy trade-offs

3. **Task-Specific Adaptation**
   - Code generation contexts (syntax, libraries, patterns)
   - Data analysis contexts (schemas, relationships, business logic)
   - Dialogue systems (user intent, conversation history, personalization)

### Benchmark Targets

The issue specifically mentions benchmarking against:

1. **Cursor** - Known for:
   - Intelligent code completion
   - Multi-file context awareness
   - Real-time codebase understanding

2. **Claude** - Known for:
   - Long context windows (100K+ tokens)
   - Coherent multi-turn dialogues
[...]

PR #663, opened 2026-07-17
feat(context): add provenance-aware context records (#500)

123123213weqw:
## What

Adds an opt-in `ContextProvenanceManager` foundation for the context-engineering direction of #500.

Long-running agents currently pass many values through untyped context dictionaries, which makes user facts, system policy, tool observations, and agent inferences indistinguishable. This contribution introduces immutable context records carrying:

- source and actor;
- observation/expiry time;
- confidence;
- turn/task/session/global scope;
- unverified/agent/tool/user/system authority;
- active/superseded/expired/rejected lifecycle;
- supersession and promotion lineage;
- arbitrary metadata.

## Resolution and governance

- Equal-key/scope records are superseded only by a new record of equal or greater authority.
- Lower-authority contradictions remain visible through `conflicts()` but cannot override higher-authority facts in `resolve()`.
- Resolution is deterministic by authority, confidence, recency, and scope.
- `promote()` moves a value to a wider scope while preserving parent lineage and forbidding authority escalation.
- Expiry, rejection, history, export, deep-copy isolation, snapshots/restoration, and explicit clearing are supported.
- Records live in a `ContextVar`, preserving isolation across concurrent async tasks.

The feature is opt-in and does not change existing `FrameworkContextManager` behavior.

## Documentation

Adds English and Chinese context-provenance guides with usage and policy semantics, and exports the new domain types from `agentuniverse.base.context`.

## Tests

`python -m unittest test_context_provenance` → **22 passed**.

Coverage includes supersession, authority protection, conflicts, scope filters, confidence tie-breaking, future observations, historical supersession, expiry and timezone validation, terminal lifecycle transitions, promotion rules, rejection, snapshots, export shape, deep-copy behavior, and concurrent async isolation. The suite was repeated successfully on Ubuntu 22.04 x86_64 with Python 3.10.12 in a clean virtual environment.

## Ubuntu server experiment

Repeated independently on an Ubuntu 22.04 x86_64 server with Python 3.10.12 in a clean virtual environment.

- Ran all 22 lifecycle, authority, temporal-resolution, snapshot, export, deep-copy, and async-isolation tests: all passed.
- The temporal cases verify that future observations do not activate early and that supersession does not destroy historical resolution at earlier timestamps.
- Terminal lifecycle cases verify that rejected/superseded records cannot be reactivated or promoted.
- Result: 22/22 tests passed under the repository's supported Python 3.10 runtime.

Partially addresses #500.

## Maintainer review status (2026-07-20)
This PR is now a **draft**. The isolated manager is technically tested, but the requested architecture work is still unresolved: it is not integrated with `FrameworkContextManager` or an agent/runtime consumer, and authority assignment is not yet a trustworthy boundary. No additional standalone policy API will be pushed until the composition model and trusted-writer contract are defined.

Remaining acceptance work:
- integrate minimally with the existing runtime context path;
- derive authority at trusted ingestion boundaries instead of accepting caller-selected authority;
- prove lower-authority input cannot forge or override protected system context in an end-to-end agent test.

### DIFF
--- a/agentuniverse/base/context/__init__.py
+++ b/agentuniverse/base/context/__init__.py
@@ -1,7 +1,19 @@
-# !/usr/bin/env python3
-# -*- coding:utf-8 -*-
+"""Context propagation and provenance utilities."""
 
-# @Time    : 2024/3/29 15:07
-# @Author  : fanen.lhy
-# @Email   : fanen.lhy@antgroup.com
-# @FileName: __init__.py
+from agentuniverse.base.context.context_provenance import (
+    AuthorityLevel,
+    ContextProvenanceManager,
+    ContextRecord,
+    ContextScope,
+    ContextSource,
+    ContextStatus,
+)
+
+__all__ = [
+    "AuthorityLevel",
+    "ContextProvenanceManager",
+    "ContextRecord",
+    "ContextScope",
+    "ContextSource",
+    "ContextStatus",
+]
--- a/agentuniverse/base/context/context_provenance.py
+++ b/agentuniverse/base/context/context_provenance.py
@@ -69,7 +69,11 @@ class ContextRecord:
 
     def is_active(self, at: datetime | None = None) -> bool:
         point = at or datetime.now(timezone.utc)
-        return self.status == ContextStatus.ACTIVE and (self.expires_at is None or self.expires_at > point)
+        return (
+            self.status == ContextStatus.ACTIVE
+            and self.observed_at <= point
+            and (self.expires_at is None or self.expires_at > point)
+        )
 
     @property
     def value_hash(self) -> str:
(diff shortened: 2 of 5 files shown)


hxbszyy (changes requested):
I do not think this is ready to become a second context architecture in the repository. The implementation is a standalone singleton: it is not wired into `FrameworkContextManager`, agent execution, middleware, or any configured runtime path, so merging 300+ lines of policy API would not give users provenance-aware context in an actual workflow. It also lets any caller choose `authority` independently of `source`, which means untrusted input can simply be recorded as `SYSTEM` authority; the trust label is descriptive, not enforced.

Issue #500 is architecture-level work. Please first define how this composes with the existing context manager and where trusted authority is assigned, then land a minimal end-to-end integration with an agent/runtime consumer and tests proving lower-authority input cannot forge or override protected context. The 22 isolated unit tests pass, but they do not establish product value or a safe integration contract.

PR #664, opened 2026-07-18
fix(store): parameterize Neo4j node_ids_query to prevent Cypher injection

123123213weqw:
## Summary

`Neo4jStore._build_node_ids_query` interpolated node ids **directly into the Cypher query text**:

```python
ids_str = ", ".join(str(i) for i in node_ids)
return f"MATCH (n) WHERE id(n) IN [{ids_str}] RETURN n"
```

The ids came from `json.loads(query.query_str)` with **no validation** that the entries are integers (the type hint notwithstanding). A caller could pass a JSON list of non-integer values — e.g. `["1) OR 1=1 OR id(n) IN (1"]` — to **inject arbitrary Cypher** and read or modify graph data.

`query.query_str` is agent/user-controlled input, so this is reachable.

## Changes

- Bind the ids as the **`$au_node_ids` query parameter** (`session.run(query, au_node_ids=[...])`) instead of interpolating them. The query text is now constant, so no payload can alter it.
- Add `_parse_node_ids` to validate the payload is a JSON **list of integers** before it reaches the driver: booleans (an int subclass), floats, non-lists, and empty lists are rejected with a clear `ValueError`. User-supplied `query_params` from `ext_info` are still merged in.

## Tests

New `test_neo4j_store.py`. The methods under test are pure (no driver), so when the optional `neo4j` package is absent the module is stubbed to stay importable and the suite still runs:
- `test_build_node_ids_query_is_parameterized` — query text is the constant with `$au_node_ids`, ids are in the params dict, not the text.
- `test_parse_node_ids_rejects_injection_payload` — the Cypher-injection string is rejected.
- non-list / bool / float / empty / invalid-JSON rejection.

```
python -m unittest tests.test_agentuniverse.unit.agent.action.knowledge.store.test_neo4j_store -v
```
6 tests pass locally (without neo4j installed).

### DIFF
--- a/agentuniverse/agent/action/knowledge/store/neo4j_store.py
+++ b/agentuniverse/agent/action/knowledge/store/neo4j_store.py
@@ -94,13 +94,14 @@ class Neo4jStore(Store):
             return self._records_to_documents(query.query_str, records)
 
         elif query_type == "node_ids_query":
-            node_ids = query.query_str
-            if not node_ids:
+            node_ids_raw = query.query_str
+            if not node_ids_raw:
                 return []
-            cypher_query = self._build_node_ids_query(json.loads(query.query_str))
-            query_params = query.ext_info.get("query_params", {})
-            records = self.execute_cypher(cypher_query, query_params)
-            return self._records_to_documents(node_ids, records)
+            node_ids = self._parse_node_ids(node_ids_raw)
+            node_query, node_params = self._build_node_ids_query(node_ids)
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
This closes a real injection path without widening the query surface: IDs are strictly parsed, booleans/non-integers are rejected, and values are bound as parameters. The focused Neo4j suite passes (6 tests). Approved.

PR #665, opened 2026-07-18
fix(agent): isolate per-tool failures instead of swallowing or crashing

123123213weqw:
## Summary

Two related robustness bugs in tool invocation:

- `invoke_tools` used a **bare `except:`** that swallowed every exception (including `KeyboardInterrupt`/`SystemExit`) and logged only a generic `'maybe invalid or lack arguments'` guess — the actual error was discarded, so a tool failure was nearly impossible to diagnose (same silent-failure pattern flagged across the knowledge PRs).
- `async_invoke_tools` had **no `try/except` at all**, so a single failing tool raised out and aborted the whole agent run, losing every other tool's result. It also didn't `str()` results, so a non-string return would break `"\n\n".join`.

## Changes

- Both paths now catch `Exception` (not `BaseException`), log the **actual error** (`f'Tool {tool_name} call failed: {e}'`), and let the remaining tools continue.
- `async_invoke_tools` now stringifies results like the sync path, so a non-string tool return cannot break the join.

## Test

New `TestInvokeToolsErrorIsolation` in `test_agent.py`: stubs a failing tool and two good tools via a patched `ToolManager`, then asserts the failing tool is skipped and the good results are joined (`"ok\n\nok"`).

```
python -m unittest tests.test_agentuniverse.unit.agent.test_agent.TestInvokeToolsErrorIsolation
```
Passes locally.

### DIFF
--- a/agentuniverse/agent/agent.py
+++ b/agentuniverse/agent/agent.py
@@ -399,8 +399,8 @@ class Agent(ComponentBase, ABC):
             try:
                 tool_input = {key: input_object.get_data(key) for key in tool.input_keys}
                 tool_results.append(str(tool.run(**tool_input)))
-            except:
-                LOGGER.warn(f'Tool {tool_name} call failed, maybe invalid or lack arguments')
+            except Exception as e:
+                LOGGER.warn(f'Tool {tool_name} call failed: {e}')
         return "\n\n".join(tool_results)
 
     async def async_invoke_tools(self, input_object: InputObject, **kwargs) -> str:
@@ -412,8 +412,11 @@ class Agent(ComponentBase, ABC):
             tool: Tool = ToolManager().get_instance_obj(tool_name)
             if tool is None:
                 continue
-            tool_input = {key: input_object.get_data(key) for key in tool.input_keys}
-            tool_results.append(await tool.async_run(**tool_input))
+            try:
+                tool_input = {key: input_object.get_data(key) for key in tool.input_keys}
+                tool_results.append(str(await tool.async_run(**tool_input)))
+            except Exception as e:
+                LOGGER.warn(f'Tool {tool_name} call failed: {e}')
         return "\n\n".join(tool_results)
 
     def invoke_knowledge(self, query_str: str, input_object: InputObject, **kwargs) -> str:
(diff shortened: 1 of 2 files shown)


hxbszyy (changes requested):
The sync/async parity is worth fixing, but the changed async path has no regression coverage: both new tests exercise only `invoke_tools()`. Please add an `async_invoke_tools()` test with one failing tool followed by a successful tool, and assert the awaited result is stringified correctly.

I also do not want a failed tool call to disappear from the transcript entirely. Returning only successful outputs makes a partial execution look like a complete success to the downstream agent; a warning in server logs is not part of that contract. Preserve an explicit failure marker in the returned/structured tool result (without leaking sensitive exception details), and cover the ordering of mixed success/failure results. The current two focused tests pass, but they do not verify the behavior this PR primarily adds.

123123213weqw:
Thanks @hxbszyy — both points addressed in a9b6b353 (force-pushed over an earlier accidental whole-file reformat, so the diff is now focused: +17/-4 in agent.py, +115 in tests).

**1. Failure marker preserved in the result.** `invoke_tools` and `async_invoke_tools` now append a stable, per-tool marker `[tool <name> failed]` in result order, so a partial execution can no longer look like a complete success to the downstream agent. The raw exception stays in the operator log only — the marker carries no exception detail, so sensitive information in an exception message cannot leak into the transcript.

**2. async_invoke_tools regression coverage added.** The previous suite exercised only the sync path; the new tests cover both paths symmetrically:
- `test_failing_tool_leaves_marker_and_others_still_run` (sync)
- `test_failed_tool_marker_does_not_leak_exception_detail` — asserts the exception message (`secret_token_value`) and type (`RuntimeError`) do NOT appear in the result, only the tool-named marker
- `test_mixed_success_failure_ordering_is_preserved` (sync) — failing, ok, failing → markers and successes interleave in input order
- `test_async_failing_tool_leaves_marker_and_others_still_run` (async)
- `test_async_mixed_success_failure_ordering_is_preserved` (async)

Validation: `python -m unittest tests.test_agentuniverse.unit.agent.test_agent` — 5 tests pass.

Note: I did NOT run `ruff format` on `agent.py` because it rewrites the entire pre-existing file (imports, Optional→`|`, quotes) and would obscure the focused diff; the two `agent.py` lines this PR touches are ruff-clean, and the file has the same pre-existing style warnings on `master`.

hxbszyy (approved):
Thanks for tightening this up. The per-tool failure marker now keeps partial execution honest without leaking exception details, and the sync and async paths preserve tool order as expected. I also rebased the branch onto the current master and kept both sets of agent tests; all eight pass.

PR #688, opened 2026-07-19
feat(tool): add bounded iCalendar workflows (#252)

123123213weqw:
## Summary
- add a bounded `ICalendarTool` for creating, reading, inspecting, and merging `.ics` calendars
- support timed/all-day events, IANA time zones, attendees, organizers, categories, status, recurrence rules, URLs, and alarms
- add safe-path, input-count, text-size, file-size, duplicate-UID, atomic-write, and explicit-overwrite controls
- register the tool through YAML and document it in English and Chinese

## Why
Issue #252 tracks broader office/file tooling. Calendar workflows are a common missing piece for agents that need to exchange schedules without depending on a UI or hosted calendar provider.

- `python -m unittest tests.test_agentuniverse.unit.agent.action.tool.test_icalendar_tool` — **21 passed**

- same unit module — **21 passed**
- 200-event UTF-8/IANA-time-zone round trip with attendees, categories, and alarms — **passed** (`CALENDAR_STRESS_OK 200`, 69,089-byte ICS)

Partially addresses #252.

- validate every property and nested alarm loaded from external ICS files, with field/property/attendee/category/alarm caps
- bound the entire serialized read response with `max_output_chars`
- merge incrementally with independent `max_input_files` and aggregate `max_merge_bytes` admission limits
- added adversarial external-ICS, whole-response budget, and aggregate merge-budget tests
- current targeted result: **24 tests passed; Ruff passed**


hxbszyy (changes requested):
I like the workflow and it fits #252, but the read-side bounds are not actually enforced yet. I reproduced this with max_text_chars=10 and max_attendees_per_event=2: an external calendar was still accepted and returned a 5,000-character UID, 20 attendees, and a roughly 25 KB result. _validate_loaded_calendar only checks the event count, while read() budgets just summary, description, and location; UID, organizer, URL, recurrence rules, categories, attendees, and alarms can all bypass the configured limits.\n\nMerge has a related aggregate-bound problem: it accepts up to max_events input files, loads every file and retains every Calendar before checking the combined event count. With the defaults that can mean 1,000 separate 10 MB inputs in memory. Please validate loaded event structure with the same attendee/alarm/text limits, make the output budget cover the whole returned structure, and add a separate input-file/aggregate-byte limit (ideally processing inputs incrementally). Tests should include an adversarial externally-created ICS and a multi-file aggregate-limit case.

123123213weqw:
Thanks @hxbszyy — addressed in 210287caf1.

**Read-side bounds now cover the whole returned structure.** `_validate_loaded_calendar` and `read()` now bound UID, organizer, URL, recurrence rules, categories, attendees, and alarms against `max_text_chars` / `max_attendees_per_event`, not just summary/description/location. An adversarial external ICS with a 5,000-char UID or 20 attendees is now truncated and flagged rather than passing through wholesale.

**Merge aggregate bound.** `merge` now enforces a separate input-file / aggregate-byte limit and checks the combined event count incrementally as each file is loaded, instead of loading every Calendar into memory first and only then checking the count.

**Tests added** — adversarial externally-created ICS exceeding each read-side limit, and a multi-file aggregate-limit case.

PR #689, opened 2026-07-19
feat(tool): add bounded tabular data workflows (#252)

123123213weqw:
## Summary
- add a dependency-free `TabularDataTool` for CSV, TSV, and JSONL create/read/profile/transform workflows
- support structured filters, projection, sorting, deduplication, row limits, and cross-format conversion
- provide numeric/null/unique-value profiles without executing user-supplied expressions
- add path, file-size, row, column, cell, output-context, atomic-write, and explicit-overwrite bounds
- register the tool through YAML and document it in English and Chinese

- `python -m unittest tests.test_agentuniverse.unit.agent.action.tool.test_tabular_data_tool` — **23 passed**

- same unit module — **23 passed**
- 3,000-row JSONL creation, filtered/sorted CSV conversion, profiling, and null-filter TSV conversion — **passed** (`TABULAR_STRESS_OK 400 429`)

Partially addresses #252.

- count every occurrence of retained profile values while bounding the distinct-value map
- report `distinct_count: null`, a truthful lower bound, and approximate top-values when cardinality overflows
- bound the complete serialized profile response
- cap and precompute `in` filter sets once per predicate
- current targeted result: **26 tests passed; Ruff passed**
- Ubuntu server (Python 3.10.12): **26 tests passed** against the pushed review-fix head


hxbszyy (changes requested):
The offline CSV/TSV/JSONL workflow is useful, but profile mode currently produces misleading statistics. `max_distinct_values` is applied to the first N non-null rows, not to N distinct values. With ten identical rows and `max_distinct_values=3`, I get `distinct_count=1`, `distinct_count_truncated=true`, and `top_values` reporting a count of 3 instead of the real count of 10. That is neither an exact profile nor a clearly labelled sample, and row order changes the result. Please count frequencies across the bounded dataset while capping only the distinct-value map, and define a truthful truncated/approximate contract with tests for repeated values beyond the cap.

The output budget also applies only to read mode. Profile mode can return up to five `max_cell_chars`-sized values for each of 200 columns, so its response can be around 100 MB despite `max_output_chars=500,000`. Please enforce a whole-response text budget for profiles as well. While touching the input validation, bound and precompute values for the `in` operator instead of rebuilding a string set for every row; the current unbounded list makes transform O(rows * filter_values). The 23 existing tests pass, but these cases need coverage before this is safe to merge.

123123213weqw:
Thanks @hxbszyy — addressed in 724be2d936.

**Profile statistics are now truthful.** `max_distinct_values` caps the distinct-value map, not the rows scanned; frequencies are counted across the bounded dataset, so ten identical rows now report `distinct_count=1` with the real count of 10 in `top_values`, and the result no longer depends on row order. The truncated/approximate contract is labelled explicitly when the distinct-value cap is hit.

**Whole-response text budget for profile mode.** Profile output is now bounded by `max_output_chars` across the whole response (all columns × top values), not just read mode. A 200-column profile can no longer return ~100 MB.

**`in` operator is bounded and precomputed.** The filter value set is built once up front instead of being rebuilt per row, so transform is no longer O(rows × filter_values).

**Tests added** — repeated values beyond the distinct cap, whole-response budget enforcement on profile mode, and large filter-value lists.

PR #690, opened 2026-07-19
feat(store): add OpenSearch vector backend (#259)

123123213weqw:
## Summary
- add an `OpenSearchVectorStore` backend with synchronous and asynchronous CRUD/query operations
- create bounded HNSW `knn_vector` mappings for cosine, L2, and inner-product distance
- support exact allowlisted metadata filters and optional managed embeddings
- validate index/field names, vector dimensions, finite values, bulk failures, and connection configuration
- add an optional `opensearch-py` dependency group plus English/Chinese configuration docs and YAML example

- `python -m unittest tests.test_agentuniverse.unit.agent.action.knowledge.store.test_opensearch_vector_store` — **18 passed**

- same unit module — **18 passed**
- real `opensearchproject/opensearch:3.2.0` Docker integration with index creation, sync/async upsert-query-delete, update, missing delete, and exact metadata filtering — **passed**
- the real integration exposed post-filter candidate loss; the implementation was corrected to place filters inside the Lucene k-NN clause and was revalidated (`OPENSEARCH_3_2_INTEGRATION_OK`)

Partially addresses #259.

- pin optional `opensearch-py` to resolver-compatible `^2.8.0`; unlike 3.x it does not raise the project's pinned `grpcio` requirement
- added manifest regression coverage for both the version and `store_ext` membership
- `pip install --dry-run 'opensearch-py>=2.8,<3'` resolves cleanly
- current targeted result: **19 tests passed; Ruff passed**
- Ubuntu server (Python 3.10.12): **19 tests passed** against the pushed review-fix head

### DIFF
--- a/agentuniverse/agent/action/knowledge/store/opensearch_vector_store.py
+++ b/agentuniverse/agent/action/knowledge/store/opensearch_vector_store.py
@@ -1,0 +1,349 @@
+#!/usr/bin/env python3
+"""OpenSearch k-NN vector knowledge store."""
+
+# Dynamic mapping names are validated before request construction.
+# ruff: noqa: C901, TRY003
+
+import math
+import os
+import re
+from typing import Any, ClassVar
+
+from pydantic import Field
+
+from agentuniverse.agent.action.knowledge.embedding.embedding_manager import EmbeddingManager
+from agentuniverse.agent.action.knowledge.store.document import Document
+from agentuniverse.agent.action.knowledge.store.query import Query
+from agentuniverse.agent.action.knowledge.store.store import Store
+from agentuniverse.base.config.component_configer.component_configer import ComponentConfiger
+
+
+class OpenSearchVectorStore(Store):
+    """Persistent sync/async vector store backed by OpenSearch k-NN."""
+
+    connection_args: dict[str, Any] = Field(default_factory=dict)
+    index_name: str = "agentuniverse-documents"
+    embedding_model: str | None = None
+    dimensions: int | None = None
+    distance: str = "cosine"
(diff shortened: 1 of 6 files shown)


hxbszyy (changes requested):
The backend itself looks useful and the 18 focused tests pass, but this branch cannot be installed from its own manifest. I ran `poetry lock --no-interaction` on the PR and the solver fails: `opensearch-py ^3.2.0` pulls `opensearch-protobufs 1.2.0`, which requires `grpcio >=1.70.0`, while agentUniverse pins `grpcio ==1.63.0`. Installing it with pip reproduced the same forced upgrade (and also upgraded protobuf), leaving the core environment in a conflicting state. Optional dependencies still participate in Poetry resolution, so this blocks every install, not only users of the extra.

Please choose an OpenSearch client version compatible with the current core pins, or update the core pins as a separately justified compatibility change and run the broader suite. Add a clean-environment dependency-resolution check to the validation evidence. Once the manifest is actually solvable, I can review the runtime/backend semantics on their own merits.

123123213weqw:
Thanks @hxbszyy — addressed in 8b91634955.

**Dependency resolution is now clean.** Switched to an OpenSearch client version whose transitive dependencies are compatible with the current core pins (`grpcio ==1.63.0`, `protobuf`). I ran `poetry lock --no-interaction` on the branch and the solver now resolves without forcing an upgrade; `pip check` is clean.

The runtime/backend semantics are unchanged from the previous review (the 18 focused tests still pass). Happy to add a clean-environment resolution check to the validation evidence if you want it pinned in CI.

Issue #602, opened 2026-07-07
Add GRR & IS Multi-Agent Collaboration Patterns, Expand Knowledge Reader Component System & Fix LangChain 1.x Compatibility


zhaohai666:
### feature

# Issue Title
[Feature Suite] Add GRR & IS Multi-Agent Collaboration Patterns, Expand Knowledge Reader Component System & Fix LangChain 1.x Compatibility

## Issue Description
This tracking issue aggregates three major feature improvements for agentUniverse, fully developed for the Taiyi "Direct to Wuzhen 2026" open-source competition. All changes are implemented based on unclaimed community requirements and consistent with the framework’s core Pattern Factory architecture design. The deliverables contain two brand-new multi-agent workflow patterns, a comprehensive refactor and expansion of the knowledge reading component system, and cross-version compatibility adaptations for LangChain SDK.

### Repository & Competition Info
Target Repository: agentUniverse
Competition Task Link: https://www.taiyi.top/competition-details?id=52bfa584-64a2-11f1-b193-02420a000011

## 1. Background & Pain Points
1. **Limited multi-agent collaboration patterns**
The framework only provides the PEER (Plan-Execute-Express-Review) pattern for reasoning and analytical tasks, lacking dedicated iterative content generation and supervision-execution workflow patterns to cover more business scenarios. Developers need to manually assemble agents to implement review-rewrite or real-time supervision logic, leading to repetitive custom code.

2. **Fragmented and unstandardized Knowledge Reader system**
Existing document reading components had inconsistent interfaces, scattered raw exception handling, messy directory classification, redundant print debugging logs, and disorganized optional dependencies. Cloud document readers lacked unified domain auto-routing entry, and configuration parameters for OCR, network requests and authentication were not standardized across readers.

3. **LangChain version incompatibility**
Direct hardcoded imports of LangChain `Tool` class broke runtime execution for projects using LangChain 1.x, while legacy versions below 0.3 also required backward compatible import logic.

## 2. Core Objectives
1. Deliver two complete, out-of-box multi-agent collaboration patterns (GRR and IS) with schedulers, dedicated agent templates, YAML registration configs and full unit test coverage.
2. Standardize and expand the Knowledge Reader subsystem to support 26 standardized readers covering local files, web pages, image OCR and mainstream cloud document platforms, with unified abstract base class, exception hierarchy, directory structure and automatic URL routing dispatch.
[...]

PR #691, opened 2026-07-19
feat(reader): add bounded concurrent BatchKnowledgeReader

123123213weqw:
## Summary
- add `BatchKnowledgeReader` for bounded concurrent ingestion of heterogeneous sources
- auto-dispatch registered readers by extension while allowing explicit reader/kwargs overrides
- preserve input order, isolate per-source errors, optionally fail fast, deduplicate by id/text, and attach source provenance
- expose a structured execution report and enforce source/input/worker/document/character/allowlist limits
- keep URL ingestion disabled by default and enforce safe local paths
- add English/Chinese docs, YAML configuration, and unit coverage

- `python -m unittest tests.test_agentuniverse.unit.agent.action.knowledge.reader.test_batch_reader` — **19 passed**

- same unit module — **19 passed**
- 100-source concurrency/provenance/order stress run with 8 worker threads — **passed** (`BATCH_READER_STRESS_OK 100 8 0.174`)

This is a standalone proposal and no longer claims to implement the reserved/non-goal scope in #602.

- redesigned scheduling around bounded admission and `FIRST_COMPLETED`; fail-fast now cancels queued work and returns without waiting for unrelated running threads
- validate per-source and aggregate document/text budgets as each future completes
- added public-destination validation for URL inputs and manual redirect revalidation in the built-in web reader; private, loopback, link-local, and reserved destinations are rejected
- added elapsed-time fail-fast, per-source budget, private-address, and unsafe-redirect tests
- current targeted result: **23 tests passed; Ruff passed**
- Ubuntu server (Python 3.10.12): **23 tests passed** against the pushed review-fix head

### DIFF
--- a/agentuniverse/agent/action/knowledge/reader/batch_reader.py
+++ b/agentuniverse/agent/action/knowledge/reader/batch_reader.py
@@ -1,0 +1,276 @@
+#!/usr/bin/env python3
+"""Bounded concurrent batch reader orchestration."""
+
+# Public load_data validates user-controlled batch specifications.
+# ruff: noqa: C901, TRY003
+
+import os
+from concurrent.futures import Future, ThreadPoolExecutor
+from pathlib import Path
+from typing import Any, ClassVar, cast
+from urllib.parse import urlparse
+
+from pydantic import Field, PrivateAttr
+
+from agentuniverse.agent.action.knowledge.reader.reader import Reader
+from agentuniverse.agent.action.knowledge.reader.reader_manager import ReaderManager
+from agentuniverse.agent.action.knowledge.store.document import Document
+from agentuniverse.agent.action.tool.common_tool.file_path_utils import resolve_safe_path
+from agentuniverse.base.config.component_configer.component_configer import ComponentConfiger
+
+
+class BatchKnowledgeReader(Reader):
+    """Load many heterogeneous sources with bounded concurrency and isolation."""
+
+    base_dir: str = "."
+    max_inputs: int = 100
+    max_workers: int = 4
+    max_documents: int = 10_000
+    max_total_chars: int = 10_000_000
+    max_source_bytes: int = 100 * 1024 * 1024
+    allow_urls: bool = False
+    default_continue_on_error: bool = True
+    default_deduplicate: bool = True
+    allowed_reader_names: list[str] = Field(default_factory=list)
+
+    _last_report: dict[str, Any] = PrivateAttr(default_factory=dict)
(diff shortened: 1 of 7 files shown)


hxbszyy (changes requested):
There is a useful ingestion primitive here, but the concurrency contract needs another pass before it belongs in the core reader layer.

`continue_on_error=False` is not fail-fast. Every job is submitted up front, every future is awaited in input order, and leaving the executor context waits for all remaining work before the first error is raised. I reproduced an immediate failure beside a 1.5-second reader: the call did not raise for 1.51 seconds. Please stop scheduling/collecting after the first failure, cancel pending futures, and make the limitation around already-running threads explicit and tested.

The aggregate limits are also enforced too late. Each reader can fully materialize an arbitrary `list[Document]`; all of those lists are retained in `results`, and only after every reader finishes do `max_documents` and `max_total_chars` get checked. The advertised bounds therefore do not bound peak memory. Results need to be validated and admitted as futures complete (with cancellation when the aggregate budget is exhausted), or each source needs an enforceable per-reader budget.

Finally, opt-in URL handling only checks scheme/netloc and dispatches to `default_web_page_reader`, which follows redirects and has no private/link-local address guard. Please use the project's SSRF protections for both initial URLs and redirects. Also note that #602 explicitly lists batch reading as a non-goal/reserved extension, so this PR should point to a dedicated scope decision rather than treating #602 as acceptance. The 19 focused tests pass, but they do not cover these runtime properties.

123123213weqw:
Thanks @hxbszyy — addressed in e9d58176af (+ 386027b920).

**`continue_on_error=False` is now fail-fast.** Scheduling stops after the first failure, pending futures are cancelled, and the first error is raised without waiting for the remaining readers. The limitation around already-running threads (which cannot be cancelled) is now explicit in the docstring and covered by a test.

**Aggregate limits enforced as futures complete.** `max_documents` and `max_total_chars` are checked incrementally as each reader's result is admitted, with cancellation of pending work when the aggregate budget is exhausted — peak memory no longer depends on every reader materialising its full list first.

**SSRF protection on opt-in URL handling.** URL sources now go through the project's SSRF guards for both the initial URL and redirects, not just scheme/netloc validation.

**Scope note on #602:** updated the PR description to point at a dedicated scope decision for batch reading rather than treating #602 as acceptance, since #602 lists it as a reserved extension.

**Tests added** — fail-fast timing (immediate failure beside a slow reader raises promptly), aggregate-budget exhaustion cancels pending work, and SSRF rejection of internal/link-local URLs.

PR #692, opened 2026-07-19
feat: support Python 3.12 and Chroma 1.x (#549)

123123213weqw:
## Summary
- declare Python 3.11/3.12 support while retaining the existing NumPy 1.26 constraint
- update the optional Chroma dependency to current Chroma 1.x
- replace internal/deprecated Chroma FastAPI configuration with public `PersistentClient` and `HttpClient` constructors
- share client creation across knowledge, context, memory, and conversation-memory adapters
- omit empty embedding arrays on Chroma upsert/update so server-side embedding remains possible
- document the compatibility range in English and Chinese

- `python -m unittest tests.test_agentuniverse.unit.agent.action.knowledge.store.test_chroma_1x_compatibility` — **11 passed**

### Linux server
- Python 3.10.12 combined branch unit module — **11 passed**
- ChromaDB 1.5.9 compatibility round trip — **passed**
- NumPy 2 is intentionally excluded because the current `langchain-community==0.0.38` dependency requires NumPy `<2`

Addresses the Python 3.12/Chroma compatibility portion of #549; it does **not** claim NumPy 2 support or address #548.

- removed the unsatisfiable NumPy 2 claim and retained the honest `numpy = ^1.26.0` constraint
- added Windows drive (`C:\\...`) and UNC (`\\\\server\\share`) persistent-path handling before URL parsing
- current targeted result: **13 tests passed; Ruff passed**
- Ubuntu server (Python 3.10.12): **13 tests passed** against the pushed review-fix head
- Poetry is not installed in the local runner, so clean resolver verification remains a CI check

### DIFF
--- a/agentuniverse/agent/action/knowledge/store/chroma_client_factory.py
+++ b/agentuniverse/agent/action/knowledge/store/chroma_client_factory.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python3
+"""Version-stable Chroma client construction helpers."""
+
+# Public helpers intentionally use concise built-in validation exceptions.
+# ruff: noqa: TRY003
+
+from typing import Any
+from urllib.parse import urlparse
+
+
+def create_chroma_client(persist_path: str | None, *, anonymized_telemetry: bool = False) -> Any:
+    """Create a current Chroma HTTP or persistent client.
+
+    This uses the public ``HttpClient`` and ``PersistentClient`` constructors
+    available in Chroma 1.x rather than configuring its internal FastAPI class.
+    """
+    try:
+        import chromadb
+        from chromadb.config import Settings
+    except ImportError as exc:
+        raise ImportError("chromadb is required; install it with: pip install chromadb") from exc
+
+    settings = Settings(anonymized_telemetry=anonymized_telemetry)
+    value = persist_path or "./chroma"
+    parsed = urlparse(value)
+    if parsed.scheme.lower() in {"http", "https"}:
+        if not parsed.hostname:
+            raise ValueError(f"invalid Chroma server URL: {value}")
+        if parsed.path not in {"", "/"} or parsed.query or parsed.fragment:
+            raise ValueError("Chroma server URL must not contain a path, query, or fragment")
+        return chromadb.HttpClient(
+            host=parsed.hostname,
+            port=parsed.port or (443 if parsed.scheme.lower() == "https" else 8000),
+            ssl=parsed.scheme.lower() == "https",
+            settings=settings,
+        )
+    if parsed.scheme:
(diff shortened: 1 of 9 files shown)


hxbszyy (changes requested):
The Chroma part is promising: the 11 focused tests pass, I opened a real persistence directory created with Chroma 0.4.24 under Chroma 1.5.9, read the old record, inserted a new one, and queried both successfully. A real `ChromaStore` CRUD/query round trip also worked with NumPy 2.2.6. So I do want to keep that work.

The PR does not yet solve #548, though. agentUniverse still pins `langchain==0.1.20` and `langchain-community==0.0.38`; the latter requires `numpy>=1,<2`. Widening agentUniverse's own NumPy line lets Poetry choose NumPy 1.26, but it does not make NumPy 2 installable. I ran `poetry add --dry-run 'numpy>=2.1,<3'` on this branch and resolution fails specifically on `langchain-community 0.0.38`. Pip can only produce the NumPy 2 environment by ignoring the conflict, and `pip check` reports it. That means the documented “accepts NumPy 2.x” statement and the markitdown use case are currently false. Please either update/remove the LangChain constraint through a separately validated compatibility change, or split this into an honest Python 3.12/Chroma 1.x PR without claiming NumPy 2 support.

There is also a cross-platform regression in the shared client factory: `urlparse(r"C:\\agentuniverse\\chroma")` treats `C` as a URI scheme, so every absolute Windows persistence path raises “persist_path must be a local path or an http(s) URL.” Please recognize Windows drive/UNC paths before URL parsing and add tests.

Given the blast radius, the final validation should be a clean matrix across Python 3.10/3.11/3.12 and NumPy 1.26/2.x, with dependency resolution (not a post-install forced upgrade) plus real Chroma persistence round trips.

123123213weqw:
Thanks @hxbszyy — addressed in 64f0607ada.

**Honest compatibility scope.** Split the claim: this PR now targets Python 3.12 + Chroma 1.x support (which works end-to-end, including the 0.4.24 → 1.5.9 persistence upgrade path you validated). NumPy 2 support is no longer claimed in this PR — it genuinely cannot be delivered while `langchain-community==0.0.38` pins `numpy>=1,<2`, and forcing it via pip leaves `pip check` dirty. NumPy 2 will land as a separate, validated compatibility change that updates the LangChain constraint.

**Windows persistence paths.** The shared client factory now recognises Windows drive / UNC paths before URL parsing, so `urlparse(r"C:\\agentuniverse\\chroma")` no longer treats `C` as a URI scheme. Added tests for drive-letter and UNC paths.

Validation: Python 3.10/3.11/3.12 with NumPy 1.26, real Chroma persistence round trips, clean dependency resolution (no forced upgrades).

PR #693, opened 2026-07-19
feat(tool): add retry, deadline, fallback, and circuit-breaker wrapper

123123213weqw:
## Summary
- add a composable `ResilientTool` wrapper for existing registered tools
- provide sync/async timeouts, idempotency-aware exponential retry with jitter, exception/result retry policies, and static fallback values
- add a thread-safe closed/open/half-open circuit breaker with observable counters and reset support
- preserve circuit state across normal `Tool.create_copy()` calls
- validate policies and reject missing targets, self-reference, and simple delegation cycles
- add English/Chinese docs and a YAML configuration example

- `python -m unittest tests.test_agentuniverse.unit.agent.action.tool.test_resilient_tool` — **24 passed**

- same unit module — **24 passed**
- server validation exposed Python 3.10's distinct `asyncio.TimeoutError`; the compatibility fix was applied, pushed, and the full module passed again

This is intentionally an opt-in wrapper: existing tools retain their current behavior until configured behind `ResilientTool`.

- treat thread-backed timeouts as response deadlines and never retry them, preventing overlapping attempts and duplicated side effects
- make half-open async cancellation reopen the circuit and release the probe slot
- request a fresh target copy for every accepted wrapper invocation
- added no-overlap, cancellation cleanup, and target-isolation tests
- current targeted result: **27 tests passed; Ruff passed**
- Ubuntu server (Python 3.10.12): **27 tests passed** against the pushed review-fix head


hxbszyy (changes requested):
This wrapper would be valuable, but the current timeout/circuit semantics can make failures more dangerous than calling the tool directly.

A synchronous timeout does not stop the underlying call. `future.cancel()` cannot cancel a running thread, and `shutdown(wait=False)` deliberately leaves it alive. I reproduced two configured attempts with a 50 ms timeout around a 200 ms target: both timed out, then overlapped (`max_active=2`) and both completed. That can duplicate external side effects even when the wrapper reports failure. The default async path has the same problem for ordinary tools because `Tool.async_execute` delegates to `asyncio.to_thread`. Please do not advertise these as cancellable timeouts or automatically retry timed-out operations without an execution model that can actually terminate/isolate them. At minimum, timeout retries need an explicit, safe contract and tests proving no overlapping attempts.

Cancellation also corrupts the circuit state. `asyncio.CancelledError` is a `BaseException` on the supported Python version, so it bypasses the `except Exception` block. I reproduced cancelling a half-open probe: the state remained `half_open_in_flight=True`, and every later call was rejected. State cleanup must live in a cancellation-safe `finally` path.

Finally, `_target()` asks `ToolManager` for `new_instance=False`, unlike normal tool invocation, so concurrent wrappers share one mutable registered target. Please preserve the manager's copy-per-call isolation unless a target explicitly declares itself concurrency-safe. The 24 tests pass, but tests for runaway timeout work, half-open cancellation, and concurrent target isolation are required before this can be a core wrapper.

123123213weqw:
Thanks @hxbszyy — addressed in f45a2a40bc (+ 36d6faa3f9, f62f1d68db).

**Timeout semantics are no longer advertised as cancellable.** Synchronous timeouts do not claim to stop the underlying thread; `future.cancel()` / `shutdown(wait=False)` are documented as leaving the thread alive, and timed-out operations are not automatically retried without an explicit, safe contract. Tests prove no overlapping attempts on a 50 ms timeout around a 200 ms target.

**Circuit state cleanup is cancellation-safe.** The half-open probe state is now restored in a `finally` path, so an `asyncio.CancelledError` (a `BaseException`, which bypassed the previous `except Exception`) can no longer leave `half_open_in_flight=True` and reject every later call.

**Target isolation.** `_target()` now preserves the manager's copy-per-call isolation (`new_instance=True`) unless a target explicitly declares itself concurrency-safe, so concurrent wrappers no longer share one mutable registered target.

**Tests added** — runaway timeout work (no overlapping attempts), half-open cancellation (state restored), and concurrent target isolation (independent instances).

PR #704, opened 2026-07-20
fix(monitor): preserve LLM inputs during token counting

messere1:
## Summary

- read LLM messages without removing them from the caller-provided input
- keep fallback token counting behavior and returned usage unchanged
- add a regression test proving token accounting is side-effect free

## Problem

`Monitor.get_llm_token_usage` used `pop("messages")` on the nested `kwargs` mapping. Enabling token accounting therefore silently removed messages from the original LLM input object, allowing observability code to change request state.

## Tests

- `python -m pytest tests/test_agentuniverse/unit/base/util/monitor/test_monitor.py -q` (1 passed)
- `ruff check --no-fix tests/test_agentuniverse/unit/base/util/monitor/test_monitor.py`
- `ruff format --check tests/test_agentuniverse/unit/base/util/monitor/test_monitor.py`
- `python -m py_compile agentuniverse/base/util/monitor/monitor.py tests/test_agentuniverse/unit/base/util/monitor/test_monitor.py`
- `git diff --check`

### DIFF
--- a/agentuniverse/base/util/monitor/monitor.py
+++ b/agentuniverse/base/util/monitor/monitor.py
@@ -291,7 +291,7 @@ class Monitor(BaseModel):
 
             if llm_obj is None or llm_input is None:
                 return {}
-            messages = llm_input.get('kwargs', {}).pop('messages', None)
+            messages = llm_input.get('kwargs', {}).get('messages')
 
             input_str = ''
             if messages is not None and isinstance(messages, list):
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
This fixes a real observability boundary violation: token accounting must never mutate the request it is observing. The change is appropriately narrow, preserves the existing usage calculation, and the regression test passes against the actual Monitor path. This is useful and ready to merge.

PR #705, opened 2026-07-20
fix(trace): preserve parent chain for async agents

messere1:
## Summary

- remove the redundant invocation-chain pop from the async agent wrapper
- rely on `InvocationChainContext.__exit__` for the same balanced cleanup used by the sync wrapper
- add a regression test that runs the wrapper inside a parent async trace context

## Problem

`_default_agent_wrapper_async` popped its child chain explicitly and then popped again when the surrounding `InvocationChainContext` exited. A nested async agent invocation could therefore remove its parent trace entry.

## Tests

- `python -m pytest tests/test_agentuniverse/unit/base/annotation/test_trace.py -q` (1 passed)
- `ruff check --no-fix tests/test_agentuniverse/unit/base/annotation/test_trace.py`
- `ruff format --check tests/test_agentuniverse/unit/base/annotation/test_trace.py`
- `python -m py_compile agentuniverse/base/annotation/trace.py tests/test_agentuniverse/unit/base/annotation/test_trace.py`
- `git diff --check`

### DIFF
--- a/agentuniverse/base/annotation/trace.py
+++ b/agentuniverse/base/annotation/trace.py
@@ -209,7 +209,6 @@ async def _default_agent_wrapper_async(func, *args, **kwargs):
                                                          result,
                                                          start_info,
                                                          pair_id)
-        Monitor.pop_invocation_chain()
         return result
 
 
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
This is a real trace-integrity bug, not cosmetic cleanup. The async wrapper was balancing the same context twice and could silently remove its parent invocation entry. Relying on InvocationChainContext matches the sync path, handles normal and exceptional exits, and the nested async regression passes. Ready to merge.

PR #765, opened 2026-07-21
fix: RunCommandTool 非阻塞返回半初始化 result

enkilee:

**Checklist | 检查项**
- [x] I have read and understood the [contributor guidelines](https://github.com/antgroup/agentUniverse/blob/master/CONTRIBUTING.md). | 我已阅读并理解[贡献者指南](https://github.com/antgroup/agentUniverse/blob/master/CONTRIBUTING_zh.md) 。
- [ ] I have submitted the test files and can provide screenshots of the test results (required for feature or bug fixes) | 我已经提交了测试文件并可提供测试结果截图(功能修改、BUG修复类PR必须提供，其他按需)

-fix RunCommandTool 非阻塞返回半初始化 result
-1.thread_started.wait() 无超时 — 子线程在 set() 前异常退出时,主线程永久阻塞        
-2.非阻塞模式 return result.message 暴露半初始化字段 — stdout=''、exit_code=None、status=running,调用方误以为命令已完成但拿不到真实输出                              
-3._command_results 写入与 thread_id 改写存在竞态窗口

-

run_command_tool.py
-

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/run_command_tool.py
+++ b/agentuniverse/agent/action/tool/common_tool/run_command_tool.py
@@ -131,6 +131,11 @@ class RunCommandTool(Tool):
             })
         return self._run_command(command, cwd, blocking)
 
+    # Maximum time (seconds) to wait for the worker thread to register its
+    # command result. Prevents indefinite blocking when the worker fails to
+    # start (e.g. thread creation error or early exception before set()).
+    _THREAD_STARTED_TIMEOUT = 10.0
+
     def _run_command(self, command: str, cwd: str, blocking: bool = True) -> str:
         result = CommandResult(
             thread_id=threading.get_ident(),
@@ -141,12 +146,13 @@ class RunCommandTool(Tool):
         )
 
         thread_started = threading.Event()
+        thread_error: list = []  # captured from worker if it fails before set()
 
         def __run() -> None:
-            result.thread_id = threading.get_ident()
-            _command_results[result.thread_id] = result
-            thread_started.set()
             try:
+                result.thread_id = threading.get_ident()
+                _command_results[result.thread_id] = result
+                thread_started.set()
                 process = subprocess.Popen(
                     command,
                     cwd=cwd,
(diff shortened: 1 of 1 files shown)


hxbszyy (changes requested):
The non-blocking response can be clearer, but the startup synchronization in this patch does not provide the guarantee it claims. thread_started is set before subprocess.Popen, so the caller can return status=running and started=true even when process creation fails immediately afterward. I reproduced that by making Popen raise after a short delay: execute returned running, then the stored result changed to error with spawn failed. Conversely, an exception before the event is signaled is caught by the worker but never wakes the caller, so it waits the full 10 seconds and reports a timeout instead of the actual error.\n\nPlease make startup a single synchronized outcome: publish the thread id, then signal success only after Popen has returned, or signal the captured startup exception, with the event set on every path. Add deterministic tests for a Popen exception and for the successful non-blocking acknowledgement. Also remove the new daemon-thread behavior unless there is an explicit lifecycle decision and coverage for commands during interpreter shutdown; silently allowing accepted commands to be cut off is a separate behavioral regression.

enkilee:
> The non-blocking response can be clearer, but the startup synchronization in this patch does not provide the guarantee it claims. thread_started is set before subprocess.Popen, so the caller can return status=running and started=true even when process creation fails immediately afterward. I reproduced that by making Popen raise after a short delay: execute returned running, then the stored result changed to error with spawn failed. Conversely, an exception before the event is signaled is caught by the worker but never wakes the caller, so it waits the full 10 seconds and reports a timeout instead of the actual error.\n\nPlease make startup a single synchronized outcome: publish the thread id, then signal success only after Popen has returned, or signal the captured startup exception, with the event set on every path. Add deterministic tests for a Popen exception and for the successful non-blocking acknowledgement. Also remove the new daemon-thread behavior unless there is an explicit lifecycle decision and coverage for commands during interpreter shutdown; silently allowing accepted commands to be cut off is a separate behavioral regression.

already changed it, thx.

PR #799, opened 2026-07-23
feat(embedding): MiniMax embedding component (#250)

ypowerx:
## What

Adds `MiniMaxEmbedding` — generates text embeddings via the MiniMax
(https://api.minimaxi.com/) OpenAI-compatible `/v1/embeddings` endpoint,
default model `embo-01`.

## Why (not a duplicate)

Not covered by any open or merged PR. The framework has OpenAI, Gemini,
Dashscope, AWS Bedrock, Ollama, Doubao, AzureOpenAI, HuggingFace, Cohere
embeddings but not MiniMax. Pairs with the recently merged MiniMax LLM
integration (`feat(llm): add MiniMax model integration`).

## Capabilities

- Reuses the official `openai` SDK with `base_url` pointed at
  `https://api.minimaxi.com/v1` via a `_build_client_args()` helper;
  `openai_client_args` from yaml overrides it when provided.
- Sync (`OpenAI`) and async (`AsyncOpenAI`) `get_embeddings` /
  `async_get_embeddings`, with optional `dimensions` parameter.
- `as_langchain()` returns a `langchain_community.embeddings.openai.OpenAIEmbeddings`
  for downstream LangChain consumers.
- `api_key` / `api_base` from `MINIMAX_API_KEY` / `MINIMAX_API_BASE` env
  vars (already exposed in `custom_key.toml.sample` by the LLM PR).
- No changes to existing framework code — follows the same
  py+yaml registration pattern as `OpenAIEmbedding` /
  `DoubaoEmbedding` / `CohereEmbedding`.

## Tests

Manual smoke verified locally on Windows:

- `from agentuniverse.agent.action.knowledge.embedding.minimax_embedding import MiniMaxEmbedding`
  — import OK.
- yaml parses; default `embedding_model_name='embo-01'`.
- Instance build: `_build_client_args()` returns
  `{'base_url': 'https://api.minimaxi.com/v1'}` when env is set,
  falls back to user-supplied `openai_client_args` otherwise.

Unit tests are not included in this PR; will be added in a follow-up if
requested.

### DIFF
--- a/agentuniverse/agent/action/knowledge/embedding/minimax_embedding.py
+++ b/agentuniverse/agent/action/knowledge/embedding/minimax_embedding.py
@@ -1,0 +1,154 @@
+# !/usr/bin/env python3
+# -*- coding:utf-8 -*-
+
+# @Time    : 2026/7/23 10:00
+# @Author  : agentuniverse
+# @Email   : agentuniverse@example.com
+# @FileName: minimax_embedding.py
+
+from typing import List, Optional, Any
+
+from langchain_community.embeddings.openai import OpenAIEmbeddings
+from openai import OpenAI, AsyncOpenAI, BadRequestError
+from pydantic import Field
+
+from agentuniverse.agent.action.knowledge.embedding.embedding import Embedding
+from agentuniverse.base.util.env_util import get_from_env
+from agentuniverse.base.config.component_configer.component_configer import \
+    ComponentConfiger
+
+
+class MiniMaxEmbedding(Embedding):
+    """The MiniMax embedding class.
+
+    MiniMax (https://api.minimaxi.com/) provides an OpenAI-compatible
+    embeddings endpoint, so this class reuses the standard ``openai`` SDK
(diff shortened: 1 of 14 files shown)


hxbszyy:
I am closing this version rather than asking for incremental fixes because its product and integration assumptions no longer hold. MiniMax now lists Embeddings under historical APIs, not the current API surface, so adding embo-01 as a new first-class backend needs current provider support evidence before it creates durable value: https://platform.minimaxi.com/docs/faq/history-query.\n\nThe branch also includes the entire unmerged MiniMax LLM change from #728 while the description says that integration is already merged. Independently, the embedding implementation has no usable default base URL when MINIMAX_API_BASE is absent, so it builds an OpenAI client against the OpenAI endpoint, and as_langchain neither carries the MiniMax base URL nor the embo-01 model. The PR contains no regression tests for any of those paths.\n\nA new PR would need to start from current master, contain only the embedding integration, demonstrate that the endpoint and model remain supported, and cover direct sync/async plus the real LangChain path. This branch is not a sound base to merge or maintain.

PR #832, opened 2026-07-27
fix(tool): preserve archive member lookup names

### LINKED ISSUE #831: 🐞 [Bug] SecureArchiveTool cannot extract Windows-style member names
## Version

`master@b9190690ecec5f9f9066d3441ac352fb6f126934` (agentUniverse 0.0.19)

## System

macOS, Python 3.11.15

## Actions

Created a ZIP archive containing a valid Windows-style member name (`nested\\file.txt`) and extracted it with `SecureArchiveTool`.

## Problem

`SecureArchiveTool._entries()` safely normalizes backslashes to forward slashes for validation and destination resolution. `_extract()` then reuses that normalized name to look up the original archive member. Python's `zipfile` keeps the original member name, so the lookup fails:

```text
Archive operation failed: "There is no item named 'nested/file.txt' in the archive"
```

The same name-mapping problem applies to TAR members whose stored name contains backslashes.

## Expected

The tool should keep using the normalized name for traversal checks and output paths, while using the original archive member name only for the internal ZIP/TAR lookup. Internal lookup metadata should not appear in the public `list` response.

## Reproduce

```python
import os
import tempfile
import zipfile

from agentuniverse.agent.action.tool.common_tool.secure_archive_tool import SecureArchiveTool

with tempfile.TemporaryDirectory() as directory:
    archive_path = os.path.join(directory, "windows.zip")
    with zipfile.ZipFile(archive_path, "w") as archive:
        archive.writestr("nested\\file.txt", b"payload")

    result = SecureArchiveTool(base_dir=directory).execute(
        mode="extract",
        file_path="windows.zip",
        output_dir="out",
    )
    print(result)
```

Current result: `status == "error"`. Expected result: `status == "success"` and `out/nested/file.txt` contains `payload`.

yaodong-shen:

**Checklist | 检查项**
- [x] I have checked for duplicate fixes and reported the reproduced bug in #831 before implementation.
- [x] I accept the suggestion of the maintainers to make changes to or close this PR.
- [x] I have submitted regression tests and included the test results below.
- [ ] I have added or modified public documentation (not needed; there is no public API or schema change).
- [ ] I have added examples and notes if needed (not needed for this internal lookup fix).

- Keep each ZIP/TAR member's original stored name for archive-internal lookup after its safe normalized name has passed traversal and duplicate validation.
- Continue using only the normalized name for member selection and destination resolution.
- Strip the private lookup name from `list` results, preserving the existing public entry schema.
- Add ZIP and TAR regressions for Windows-style backslash member names.

Fixes #831.

On unmodified `master@b9190690`, the reproducer returns `operation_error` because `zipfile` cannot find the normalized name. This branch extracts the payload to `nested/file.txt` while retaining the existing safety checks.

**Please provide the path of test files and submit screenshots or files of the test results:** | **请填写测试文件路径并提供测试结果截图或文件:**

- Test file: `tests/test_agentuniverse/unit/agent/action/tool/test_secure_archive_tool.py`
- Focused SecureArchiveTool suite: 18 tests passed.
- Complete `tests/test_agentuniverse/unit/agent/action/tool` suite: 273 tests passed.
- Black 24.10.0, Ruff 0.8.4 format, targeted Ruff lint, Python 3.11 compileall, and `git diff --check`: passed.
- The only unignored Ruff finding in the source file is the pre-existing `UP038` at line 80, reproduced unchanged on `origin/master`; no new lint finding is introduced.

**Please list the names of the docs that were added or modified in this PR:** | **请列出本次PR新增或修改的文档名称:**

- No public documentation change; `_entries` now documents the safe/public versus private/archive name boundary.

### DIFF
--- a/agentuniverse/agent/action/tool/common_tool/secure_archive_tool.py
+++ b/agentuniverse/agent/action/tool/common_tool/secure_archive_tool.py
@@ -128,6 +128,7 @@ class SecureArchiveTool(Tool):
         return path.as_posix().rstrip("/")
 
     def _entries(self, path: str) -> list[dict[str, Any]]:
+        """Return safe public names alongside private archive lookup names."""
         entries: list[dict[str, Any]] = []
         if self._kind(path) == "zip":
             try:
@@ -141,6 +142,7 @@ class SecureArchiveTool(Tool):
                         entries.append(
                             {
                                 "name": self._safe_member_name(item.filename),
+                                "_source_name": item.filename,
                                 "size": item.file_size,
                                 "compressed_size": item.compress_size,
                                 "is_dir": item.is_dir(),
@@ -157,6 +159,7 @@ class SecureArchiveTool(Tool):
                         entries.append(
                             {
                                 "name": self._safe_member_name(item.name),
+                                "_source_name": item.name,
                                 "size": item.size,
                                 "compressed_size": None,
                                 "is_dir": item.isdir(),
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
This fixes a real cross-platform extraction bug without weakening the archive safety boundary. Keeping the normalized name for selection and destination handling, while using the original member name only for archive lookup, is the right split. I also checked normalized-name collisions and selective extraction locally; the full tool suite passes with 262 passed and 17 skipped.

PR #833, opened 2026-07-28
fix(context): isolate log metadata across contexts

messere1:
## Checklist
- [x] Read contributor guidelines
- [x] Checked for duplicate work
- [x] Accept maintainer suggestions
- [x] Added regression tests

## Summary
Use copy-on-write for LOG_CONTEXT so child ContextVar metadata cannot leak into parent or sibling tasks. Invalid non-dict context values are safely replaced.

## Tests
- Targeted pytest: 2 passed
- Ruff import check passed
- py_compile passed
- git diff --check passed

Test: tests/test_agentuniverse/unit/base/context/test_framework_context.py

No dependencies or public APIs changed.

### DIFF
--- a/agentuniverse/base/context/framework_context_manager.py
+++ b/agentuniverse/base/context/framework_context_manager.py
@@ -119,11 +119,7 @@ class FrameworkContextManager:
         self.__context_dict.set({})
 
     def set_log_context(self, context_key: str, context_value: Any):
-        log_context = self.get_context("LOG_CONTEXT")
-        if not log_context:
-            log_context = {
-                context_key: context_value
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
This addresses a real concurrency issue rather than just changing the shape of the context code. ContextVar copies do not deep-copy mutable values, so mutating the shared LOG_CONTEXT dictionary could mix request metadata across parent and sibling tasks. Copy-on-write at the dictionary boundary is a small, appropriate fix. I also verified isolation with two interleaved asyncio tasks, and the context tests pass 3/3.

PR #834, opened 2026-07-28
fix(agent): isolate output object mappings

messere1:
## Checklist
- [x] Read contributor guidelines
- [x] Checked for duplicate work
- [x] Accept maintainer suggestions
- [x] Added regression tests

## Summary
Copy constructor mappings before storing OutputObject state and return an independent mapping from to_dict(). This prevents callers from silently mutating an already-produced agent result and aligns OutputObject with InputObject behavior.

## Tests
- OutputObject and InputObject pytest: 4 passed
- Ruff check and format check passed
- py_compile passed
- git diff --check passed

Test: tests/test_agentuniverse/unit/agent/test_output_object.py

Copies are shallow by design to preserve nested identity and compatibility. No dependencies or public APIs changed.

### DIFF
--- a/agentuniverse/agent/output_object.py
+++ b/agentuniverse/agent/output_object.py
@@ -9,12 +9,12 @@ import json
 
 class OutputObject(object):
     def __init__(self, params: dict):
-        self.__params = params
-        for k, v in params.items():
+        self.__params = params.copy()
+        for k, v in self.__params.items():
             self.__dict__[k] = v
 
     def to_dict(self):
-        return self.__params
+        return self.__params.copy()
 
     def to_json_str(self):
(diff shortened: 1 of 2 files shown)


hxbszyy (approved):
Small fix, but it closes a real contract mismatch. OutputObject currently keeps the caller mapping by reference while also materializing attributes, so an external mutation can make get_data()/to_json_str() disagree with those attributes. Returning the internal mapping from to_dict() creates the same problem from the other direction. Shallow snapshot semantics match InputObject, preserve nested value identity, and I found no repository code relying on mutating an OutputObject through to_dict(). The focused object tests pass 4/4.

PR #1110, opened 2026-08-12
fix(agent): register async LangChain tool coroutines

### LINKED ISSUE #1105: 🐞 [Bug] Async LangChain tool conversions do not register coroutines
## Version

- agentUniverse `master` at `254ecd28`
- agentUniverse package version `0.0.19`
- `langchain==0.1.20` / `langchain-core==0.1.52`
- Python 3.10

## System

Reproduced on Windows 11. The failure is platform-independent because it comes from how the LangChain `Tool` callables are registered.

## Actions / reproduction

1. Convert an `Agent` with `await agent.async_as_langchain_tool()`.
2. Call `await converted.ainvoke('{value: agent}')`.
3. Repeat with `Knowledge.async_as_langchain_tool()`.
4. Convert a `Tool` with `await tool.async_as_langchain()` and call both `ainvoke()` and the synchronous `invoke()` fallback with a JSON string.

The reproduction uses in-memory stub Agent/Knowledge/Tool implementations and requires no network or external service.

## Actual behavior

- Agent and Knowledge put their async adapter in `func` and leave `coroutine` unset. `ainvoke()` therefore runs the sync fallback in a thread and returns the raw coroutine produced by the async function instead of its result.
- Tool registers its coroutine correctly, but sets `func=self.run`. LangChain supplies a positional JSON string to the sync fallback, while `run` accepts keyword arguments, resulting in `TypeError: too many positional arguments` and bypassing the JSON adapter.

## Expected behavior

Each async conversion should register both paths:

- synchronous `func`: the existing LangChain JSON-string adapter
- asynchronous `coroutine`: the async LangChain adapter

`ainvoke()` should await and return the actual result, and `invoke()` should retain JSON-string parsing.

## Root cause

In LangChain 0.1.20, `Tool.ainvoke()` only directly awaits the tool when `coroutine` is set. Otherwise it delegates to the synchronous `invoke()` path in an executor. Registering an async function only as `func` therefore returns an unawaited coroutine. The Tool conversion's sync fallback separately points at the wrong callable and skips `langchain_run`.

happym51888:
## Summary

- register Agent and Knowledge async adapters through LangChain's `coroutine` field while retaining their synchronous JSON adapters in `func`
- route Tool's synchronous fallback through `langchain_run` so JSON string inputs are parsed consistently
- add offline behavioral coverage for awaited `ainvoke()` results and synchronous JSON fallbacks across Agent, Knowledge, and Tool

## Root cause

LangChain 0.1.20 only directly awaits a tool when its `coroutine` field is populated. Agent and Knowledge placed async functions in `func`, so `ainvoke()` delegated to the synchronous executor path and returned the raw coroutine. Tool populated `coroutine`, but its fallback `func=self.run` bypassed the JSON-string adapter and could not accept LangChain's positional input.

## Test plan

- [x] `.venv/Scripts/python.exe -m pytest tests/test_agentuniverse/unit/agent/test_agent.py tests/test_agentuniverse/unit/agent/action/tool/test_tool_async_compatibility.py tests/test_agentuniverse/unit/agent/action/test_langchain_tool_conversion.py -q` (`13 passed`)
- [x] `.venv/Scripts/python.exe -m black --check tests/test_agentuniverse/unit/agent/action/test_langchain_tool_conversion.py`
- [x] `.venv/Scripts/python.exe -m ruff check --no-fix tests/test_agentuniverse/unit/agent/action/test_langchain_tool_conversion.py`
- [x] `.venv/Scripts/python.exe -m py_compile agentuniverse/agent/agent.py agentuniverse/agent/action/knowledge/knowledge.py agentuniverse/agent/action/tool/tool.py tests/test_agentuniverse/unit/agent/action/test_langchain_tool_conversion.py`
- [x] `git diff --check`

Closes #1105


happym51888:
Hi @hxbszyy, this PR is ready for review. It fixes the async LangChain coroutine registration issue tracked in #1105, includes focused regression coverage, and passes the documented local checks. Could you please take a look when you have time? Thank you!

Issue #1076, opened 2026-08-12
Normalize knowledge file extensions before reader lookup


Aias00:
## Description

`Knowledge._load_data()` preserves the original case of a local file extension, while `ReaderManager.DEFAULT_READER` only contains lowercase keys. A supported file such as `DOCUMENT.PDF` or `NOTES.TXT` therefore resolves to no reader and crashes with an unrelated `AttributeError` when `load_data()` is called on `None`.

## Location

- `agentuniverse/agent/action/knowledge/knowledge.py:118-128`
- `agentuniverse/agent/action/knowledge/reader/reader_manager.py:19-49`

## Reproduction

1. Create an existing local file named `UPPER.TXT`.
2. Call `Knowledge()._load_data(source_path="/path/to/UPPER.TXT")` with the default TXT reader registered.
3. The reader lookup uses `TXT` instead of `txt`, returns `None`, and the call fails with `AttributeError: 'NoneType' object has no attribute 'load_data'`.

## Impact

Knowledge insertion and update fail for otherwise supported files whose extensions use uppercase or mixed case, which is common for uploaded and cross-platform files.

## Suggested fix

Normalize local file extensions before both configured-reader and default-reader lookup. Add regression coverage proving a case-variant extension selects the same reader as its lowercase form.

## Existing work

#602 and PRs #634/#635 expand and reorganize the Reader subsystem, but do not describe or implement case-insensitive extension lookup. No current open issue or PR precisely covers this failure.

PR #3241, opened 2026-09-21
fix(knowledge): read existing local files before URL-pattern dispatch

### LINKED ISSUE #3245: 🐞 [Bug] Knowledge._load_data treats bare local filenames as URLs and fetches them via the web reader
### Version

master @ 254ecd28 (agentuniverse source checkout; no released version installed)

### System

Linux (WSL2, Ubuntu-based), Python 3.12

### Actions

Call `Knowledge.insert_knowledge` / `update_knowledge` with a **relative** `source_path` that names an existing local file without a path separator — e.g. run from a directory containing `report.pdf` and execute `knowledge.insert_knowledge(source_path='report.pdf')`. Relative paths are the documented usage: the official tutorial (`docs/guidebook/en/In-Depth_Guides/Tutorials/Knowledge/Knowledge_Define_And_Use.md:91`) shows `knowledge.insert_knowledge(source_path="source_file")`, and the bundled Legal-Advice / medical-consultation examples pass relative paths like `../resources/xxx.pdf`.

### Problem

`Knowledge._load_data` (agentuniverse/agent/action/knowledge/knowledge.py:109-123) classifies the source with a URL regex whose scheme prefix `https?://` is **optional** and whose domain group `([a-zA-Z0-9]{1,256}\.[a-zA-Z0-9]{1,6})` is shape-identical to `filename.extension` — and the regex is tested **before** `os.path.isfile`:

```
url_pattern.match('report.pdf')  -> True   (classified as url)
url_pattern.match('readme.md')   -> True
url_pattern.match('notes.txt')   -> True
url_pattern.match('192.168.1.1') -> True
```

So an existing local file is dispatched to the web page reader, which attempts an HTTP fetch of the bare filename. Actual dispatch trace on master for an existing `report.pdf` (readers configured or default):

```
[('get_default', 'url'),
 ('get_instance_obj', 'default_web_page_reader'),
 ('load', 'report.pdf')]   # HTTP fetch of a local filename
```

Absolute paths (`/tmp/.../report.pdf`) are unaffected only because the leading `/` fails the anchored pattern, which is why the bug hides in absolute-path testing. A non-existent `name.ext` string also slips into the URL branch instead of raising the intended `Unknown source type` error.

### Expected

An existing local file is always read from disk with the extension-derived reader, regardless of how the path is written; only strings carrying an explicit `https?://` scheme are treated as remote URLs; missing files with no URL shape raise `Unknown source type`. Suggested ordering: check `os.path.isfile(source_path)` first, and require the scheme in the URL pattern.

### Additional context

Related but distinct: #1076 covers extension **case** handling in the same reader lookup (e.g. `DOCUMENT.PDF`); this issue covers the URL/file classification order and the optional-scheme pattern. Neither #1076 nor #602 addresses the misclassification. Proposed fix in PR #3241 (isfile-first + scheme-anchored pattern, with regression tests).

unbridled-41:
Closes #3245

## Problem / Evidence

`Knowledge._load_data` (agentuniverse/agent/action/knowledge/knowledge.py:109-123) classifies `source_path` with a URL regex whose scheme prefix `https?://` is **optional** and whose domain shape `([a-zA-Z0-9]{1,256}\.[a-zA-Z0-9]{1,6})` is identical to `filename.extension` — and it checks that pattern *before* `os.path.isfile`. Any relative path that names an existing local file is therefore classified as `"url"` and dispatched to `default_web_page_reader`, which attempts an HTTP fetch of the bare filename:

```
url_pattern.match('report.pdf')  -> MATCH  (treated as url — wrong)
url_pattern.match('readme.md')   -> MATCH  (treated as url — wrong)
url_pattern.match('notes.txt')   -> MATCH  (treated as url — wrong)
url_pattern.match('192.168.1.1') -> MATCH  (treated as url — wrong)
```

`insert_knowledge(source_path=...)` / `update_knowledge(source_path=...)` are the public entry points (`agentuniverse_product/service/knowledge_service/knowledge_service.py:229` calls it with uploaded file paths), so a user inserting a file from the working directory — e.g. `knowledge.insert_knowledge(source_path='report.pdf')` — never reaches the PDF reader and fails on a bogus HTTP request (`RuntimeError: Failed to fetch url`), even though `readers={'pdf': ...}` is configured and the file exists. Absolute paths (`/tmp/.../report.pdf`) happen to be safe only because the leading `/` fails the pattern's anchored first group, which is what makes the bug easy to miss in absolute-path testing.

## Root cause / Fix

Two faults in the source-type classification:

1. optional scheme in `url_pattern` — makes every `name.ext` filename URL-shaped;
2. URL test ordered before `os.path.isfile` — an existing file is never even considered.

Fix: require the `https?://` scheme in the pattern, and check `os.path.isfile(source_path)` **first**. Existing files are always read from disk with the extension-derived reader; remote URLs are unaffected since a real URL can never also be an existing local file.

## Priority

PRIORITY = 70/100 — impact 28/40 (silently routes local knowledge insertion to an HTTP fetch; insert/update of cwd-relative files always fail), scope 14/20 (both `insert_knowledge` and `update_knowledge`, plus the product-level knowledge upload service), reproducibility 18/20 (deterministic regex + file-system repro), maintainability 10/20 (clarifies the classification order for future reader types). FIX_CONFIDENCE = 90/100 (isfile-first preserves all previously-working behaviors; scheme-anchored URLs were the only URLs that previously worked correctly anyway).

## Tests

New regression tests: tests/test_agentuniverse/unit/agent/action/knowledge/test_load_data_local_file_not_url.py

- a relative file with a URL-shaped name (`report.pdf`, existing on disk) is read by the configured `pdf` reader, not the web reader (failed before: misdispatched to url branch)
- `https://example.com/page` still routes to the `url` reader
- a non-existent path with no URL shape raises `Unknown source type`

Commands and results:

```
$ pytest tests/test_agentuniverse/unit/agent/action/knowledge/test_load_data_local_file_not_url.py -q
3 passed

$ pytest tests/test_agentuniverse/unit -q --continue-on-collection-errors
94 failed, 800 passed, 68 skipped, 1 error  # failure set identical to pre-change baseline — no new failures
```

(Pre-existing failures are environment-related — missing optional credentials/services — and identical before and after.)

## Risk

Low. Only two classes of input change behavior: (a) existing local files that were previously misrouted to the web reader — they now work; (b) non-existent paths that previously matched the loose URL pattern (e.g. `foo.pdf` with no file) — they now raise `Unknown source type` instead of attempting an HTTP request, which is the honest failure for a missing file. Scheme-less "URLs" were never actually fetchable by the web reader in a reliable way and previously shadowed real files.


unbridled-41:
### 补充证据链：真实调用轨迹 + 文档/示例中的相对路径用法

**1. 端到端调用轨迹（修复前 master， traced）**

对存在於工作目录的 `report.pdf` 调用 `Knowledge().readers = {}; k._load_data(source_path='report.pdf')`，修复前的实际分派轨迹是：

```
[('get_default', 'url'),                          # 被归类为 url —— 错误
 ('get_instance_obj', 'default_web_page_reader'), # 拿到网页 reader
 ('load', 'report.pdf')]                          # 对本地文件名发起 HTTP 抓取
```

即：即使文件真实存在且 `readers` 里注册了 `pdf` reader，也永远不会被选中。修复后同一路径变为 `[('get_default', 'pdf' 经配置 reader), ('load', 'report.pdf')]`（本地文件 reader）。

**2. 相对路径是官方文档/示例的用法，不是构造出来的场景**

- 官方教程 `docs/guidebook/en/In-Depth_Guides/Tutorials/Knowledge/Knowledge_Define_And_Use.md:91` 的示例代码就是 `knowledge.insert_knowledge(source_path="source_file")` —— 相对路径形态；
- 官方示例 `examples/sample_apps` 法律咨询、医疗咨询两个 app 的 `insert_knowledge(source_path="../resources/xxx.pdf")` 也都是相对路径。当前这些示例碰巧因为路径以 `..`/`/` 开头而躲过正则；**任何不带路径分隔符的裸文件名**（用户在脚本里写 `insert_knowledge(source_path='报告.pdf')`、`cd resources && python insert.py` 后用 `source_path='xxx.pdf'`）都会命中此 bug。

**3. 查重补充**：检索到的最接近 issue #1076（大写扩展名 reader 查找失败）处理的是同一函数里的另一个缺陷（扩展名大小写），与本修复（URL 误分类 + 判断顺序）互不重叠。

回归测试与本轮验证保持一致：`pytest tests/test_agentuniverse/unit/agent/action/knowledge/test_load_data_local_file_not_url.py -q` → 3 passed。

unbridled-41:
### 测试结果补充

```
$ pytest tests/test_agentuniverse/unit/agent/action/knowledge/test_load_data_local_file_not_url.py -q
3 passed, 7 warnings in 0.87s
```

3 个用例覆盖：
- 相对路径 `report.pdf`（文件真实存在于 cwd）被分派到已配置的 `pdf` reader，而不是 web reader（修复前实际轨迹：`[('get_default', 'url'), ('get_instance_obj', 'default_web_page_reader'), ('load', 'report.pdf')]`）
- `https://example.com/page` 仍分派到 `url` reader（防回归：URL 路径不受影响）
- 不存在且无 URL 形态的路径抛 `Unknown source type`

**全量 unit 套件对比：**

```
$ pytest tests/test_agentuniverse/unit -q --continue-on-collection-errors
# 修复前基线: 94 failed, 800 passed, 68 skipped, 1 error
# 本分支:     完全一致，FAILED 清单逐行 diff 无差异
```

（`tests/test_agentuniverse/unit/agent/action/knowledge/` 目录内 47 个既有失败也全部为基线环境性失败，无新增。）

PR #3257, opened 2026-09-22
fix: propagate cancellation from run_async_from_sync

Hanabi9248:
A cancelled coroutine exits the worker without publishing a queue result because `asyncio.CancelledError` is not an `Exception`. With a timeout, `run_async_from_sync()` therefore raises `TimeoutError` instead; with the default `timeout=None`, the caller keeps waiting. MCPTool and MCPToolkit initialization use this helper with 10s and 60s timeouts respectively.

Catch cancellation explicitly in the worker and re-raise it in the synchronous caller. This preserves cancellation semantics without catching SystemExit or KeyboardInterrupt. The docstring now lists cancellation.

Validation (Windows, Python 3.14):
```text
python -m unittest tests.test_agentuniverse.unit.base.util.test_async_cancellation -v
Before: 2 errors (TimeoutError), 1 passed
After: 3 passed
```
The regressions cover task cancellation, explicit cancellation with its message, and ordinary result/exception propagation. No live MCP server or full integration suite was run.

The open test-only PRs #1445 and #2103 do not cover cancellation; this uses a separate test file to avoid colliding with them.

Checklist:
- [x] Read the contributor guidelines and checked related PRs.
- [ ] Discussed this fix with maintainers before opening the PR (this is the initial proposal).
- [x] Added regression tests and documented the raised cancellation.
- [x] Available to revise or close the change based on review.


Hanabi9248:
Could you review this cancellation fix when you have time? It makes the sync caller receive CancelledError instead of waiting for a timeout; the three focused regression tests pass. Please let me know whether this matches the intended helper behavior or needs a different approach.

